PR #1, opened 2016-06-10
Fix example yaml in README.md

plars:
I'm starting to take a look at spread, and had some trouble with the examples in the README. It looks like they don't quite match up with what spread is expecting. I think this should fix it up.


niemeyer:
Thanks!

PR #4, opened 2016-08-11
add "ssh" backend

mvo5:
This adds a new `ssh` backend to spread. The idea behind it is that spread can run tests on systems that go provisioned by some other mechanism.

I plan to use this so that I can run the spread in the autopkgtest run. The idea is that inside the autopkgtest contains we run the spread tests against "ssh:localhost-22" with something like:
```
diff --git a/debian/tests/integrationtests b/debian/tests/integrationtests
index f8e8f32..2edb829 100644
--- a/debian/tests/integrationtests
+++ b/debian/tests/integrationtests
@@ -43,7 +43,8 @@ if [ -z "$ADT_REBOOT_MARK" ]; then
     sudo cp $GOPATH/bin/snapbuild /usr/local/bin
 fi
 
+./spread -v ssh:
 
 if [ -e "${NEEDS_REBOOT}" ]; then
     mark=$(cat "${NEEDS_REBOOT}")
diff --git a/spread.yaml b/spread.yaml
index ecf7c9c..ccc2737 100644
--- a/spread.yaml
+++ b/spread.yaml
@@ -13,6 +13,12 @@ backends:
         systems:
             - ubuntu-16.04-64-grub
             - ubuntu-16.04-32-grub
+    ssh:
+      systems:
+        - localhost-22
 
 path: $[PROJECT_PATH]

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,8 @@ Convenient full-system test (task) distribution
 [Including and excluding files](#including)  
 [Selecting which tasks to run](#selecting)  
 [LXD backend](#lxd)  
-[Linode backend](#linode)  
+[Linode backend](#linode)
+[SSH backend](#ssh)  
 [More on parallelism](#parallelism)  
 
 <a name="why"/>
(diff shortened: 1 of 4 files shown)


niemeyer:
It's now conflicting due to the recent refactorings.

This should be much easier to do now after the recent improvements landed with the qemu backend. Can you please have a quick look at it when you have some spare (hah!) time?

PR #6, opened 2016-08-19
Add a reboot/resume message for -v

mvo5:
I found it helpful to see when the machine reports (when running with -v). Mostly to understand why some tests took a bit longer (or were hanging when a reboot would not return at all). Feel free to ignore if you think this only adds clutter.


niemeyer:
We already have more messaging around it nowadays. Sorry for taking so long to review this.

PR #9, opened 2016-08-24
show how many jobs are left

mvo5:
Quick drive-by branch that shows how many jobs are still pending so that the user has an idea how long he/she has to wait.


niemeyer:
The logic is a bit racy, both in the sense it isn't locking `r` while iterating over `r.pending` and also in the sense that the logged message may go out of order. We might end up with `(3 left)` after `(2 left)`. But I'll merge and fix it.

PR #12, opened 2016-09-14
client.go: when using sudo, erase SUDO_{USER,UID,GID}

mvo5:
This ensure the environment is consistent between the linode/qemu backend and the adhoc backend. Some applications behave differently when they find SUDO_USER (and friends) in the environment.


niemeyer:
Thanks!

PR #14, opened 2016-11-08
With exporting PATH now we miss the preconfigured /snap/bin in it

morphis:
/snap/bin would normally we added from the system the test is running
on but as we override PATH now we should better add it there as well.


morphis:
@niemeyer Any comments on this? Not sure if this is the right way of doing it but using /snap/bin everywhere as prefix for snap commands is another way to solve this problem.

niemeyer:
@morphis Why not just setting PATH under the `environment` field?

niemeyer:
I don't mind this either, though. Merging.

PR #15, opened 2016-12-27
Add support for globs in systems, e.g. `systems: [-ubuntu-*-ppc64el]`

mvo5:
It would be nice if spread would support wildcards in task.yaml to exclude/include systems. 

Instead of:
```
systems: [-ubuntu-core-16-64, -ubuntu-core-16-arm-64, -ubuntu-core-16-arm-32l]
```
we could use:
```
systems: [-ubuntu-core-16-*]
```
(we have more of those combinations, we now need to exclude all ppc64el targets from certain tests that use ptys in snapd).

Attached is one possible way of doing it using filepath.Match(). This seems to match nicely with what is used inside task.yaml when shell patterns are used to exclude bits and pieces.

Happy to add a proper test (using export_test.go) if that looks like a valid direction.

### DIFF
--- a/spread/project.go
+++ b/spread/project.go
@@ -440,7 +440,7 @@ func SplitVariants(s string) (prefix string, variants []string) {
 
 var (
 	validName   = regexp.MustCompile("^[a-z0-9]+(?:[-._][a-z0-9]+)*$")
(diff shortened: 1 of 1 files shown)


niemeyer:
Thanks!

PR #16, opened 2017-01-03
show aborted tasks

mvo5:
Trivial branch that shows the aborted tasks. When looking at a snapd failure this morning I only saw "aborted tasks: 2" but it was tricky to find what exactly was aborted from the long snapd test log.

### DIFF
--- a/spread/runner.go
+++ b/spread/runner.go
@@ -974,6 +974,7 @@ func (s *stats) log() {
 	printf("Successful tasks: %d", len(s.TaskDone))
 	printf("Aborted tasks: %d", len(s.TaskAbort))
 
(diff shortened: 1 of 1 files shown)


niemeyer:
Listing aborted tasks in general hints at the wrong issue. An aborted task is something that wasn't run at all as a side effect of a problem that was actually unrelated to anything that is listed. If I understand what you're looking for with this change, we need to find a way to highlight the reason why the tasks were aborted.

niemeyer:
As an example, consider what happens when we cannot allocate machines in snapd tests: we'd see a list of hundreds of tasks. None of which would help us much in the solution.

PR #17, opened 2017-01-12
properly quote a variable in MATCH

chipaca:
We were missing quotes around a variable in MATCH, meaning you couldn't use `^` and `$` in a regexp wanting to match a multiline string.


niemeyer:
Thanks!

PR #24, opened 2017-02-16
add describe option

fgimenez:
This option allows to get additional documentation in the job list.


mvo5:
Thanks for the PR, we will not move forward with it.

PR #25, opened 2017-02-16
Adding capability to write xunit report with the task suites and results

sergiocazzolato:


sergiocazzolato:
@niemeyer This is ready for reviewing, tx

mvo5:
This can be closed until we have an abstracted logging backend and then we have N implementation of logging. Doing it using if/then/else will not scale.

PR #29, opened 2017-05-25
Supporting repeating of tasks via -repeat

sergiocazzolato:
This commit introduces a new mechanism to reexecute jobs reusing the
context.
This commit includes 2 new options:
-retries <int>: defines the number of times to retry any task to be
executed
-stop-on-fail: indicates if the retries have to be stopped once the task
fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,21 +519,14 @@ are aggregated and repeated for every task under them.
 
 <a name="reexecuting"/>
 
-## Reexecuting
+## Repeating runs
 
-Reproduce an error could be a very boring experience, and Spread has a way to
-simplify that process by reexecuting the tasks as many times is desired until
+Reproducing an error may be a very boring experience, and Spread has a way to
+simplify that process by reexecuting the tasks as many times as desired until
 the task fails.
 
-To do that there are two options `-retries`, and `-stop-on-fail` which are
-used to determine the number of time to reexecute a task and if those executions
-have to be stoped when the task has failed.
-
-The `-retries` option receives an integer which indicates the number of retries 
-to do, being 0 the default value, and -1 can be used to reexecute indefinitely.
-
-All the tasks listed to be executed will be reexercuted following the same
-conditions.
+To do that there is an option `-repeat` which receives an integer indicating 
+the number of reexecutions to do, being 0 the default value.
 
 
 <a name="passwords">
--- a/README.md
+++ b/README.md
@@ -522,17 +522,17 @@ are aggregated and repeated for every task under them.
 ## Reexecuting
 
 Reproduce an error could be a very boring experience, and Spread has a way to
-simplify that process by reexecuting the jobs as many times is desired until
-the jobs fails.
+simplify that process by reexecuting the tasks as many times is desired until
+the task fails.
 
 To do that there are two options `-retries`, and `-stop-on-fail` which are
-used to determine the number of time to reexecute a job and if those executions
-have to be stoped when the job has failed.
+used to determine the number of time to reexecute a task and if those executions
+have to be stoped when the task has failed.
 
 The `-retries` option receives an integer which indicates the number of retries 
 to do, being 0 the default value, and -1 can be used to reexecute indefinitely.
 
-All the jobs listed to be executed will be reexercuted following the same
+All the tasks listed to be executed will be reexercuted following the same
 conditions.
 
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Spread
 [Timeouts](#timeouts)  
 [Fast iterations with reuse](#reuse)  
 [Debugging](#debugging)  
-[Reexecuting](#reexecuting)
+[Repeating tasks](#repeating)
 [Passwords and usernames](#passwords)  
 [Including, excluding, and renaming files](#including)  
 [Selecting which tasks to run](#selecting)  
@@ -517,9 +517,9 @@ as a `debug-each` script at the project, backend, and suite levels, so they
 are aggregated and repeated for every task under them.
 
 
-<a name="reexecuting"/>
+<a name="repeating"/>
 
-## Repeating runs
+## Repeating tasks
 
 Reproducing an error may be a very boring experience, and Spread has a way to
 simplify that process by reexecuting the tasks as many times as desired until
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Spread
 [Timeouts](#timeouts)  
 [Fast iterations with reuse](#reuse)  
 [Debugging](#debugging)  
+[Reexecuting](#reexecuting)
 [Passwords and usernames](#passwords)  
 [Including, excluding, and renaming files](#including)  
 [Selecting which tasks to run](#selecting)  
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -516,6 +517,25 @@ as a `debug-each` script at the project, backend, and suite levels, so they
...
 
+<a name="reexecuting"/>
+
+## Reexecuting
+
+Reproduce an error could be a very boring experience, and Spread has a way to
+simplify that process by reexecuting the tasks as many times is desired until
+the task fails.
+
+To do that there are two options `-retries`, and `-stop-on-fail` which are
```
niemeyer: Retrying seems to have the meaning of _trying again if it fails_ but what we mean here is _run again until it fails_, so something like _-repeat 5_ would be more clear.

I also suggest getting rid of _-stop-on-fail_ and making that the default behavior. We don't want to watch the same task break 5 different times, perhaps in different ways. Every single failure is important and should be addressed.

cmd/spread/main.go
```diff
@@ -32,6 +32,8 @@ var (
 	discard     = flag.Bool("discard", false, "Discard reused servers without running")
 	residue     = flag.String("residue", "", "Where to store residual data from tasks")
 	seed        = flag.Int64("seed", 0, "Seed for job order permutation")
+	retries     = flag.Int("retries", 0, "The number of reexecutions for each task")
```
niemeyer:
```
repeat = flag.Int("repeat", 0, "Number of times to repeat each task")

File: README.md [hunk 3]
```diff
@@ -516,6 +517,25 @@ as a `debug-each` script at the project, backend, and suite levels, so they
 are aggregated and repeated for every task under them.
 
 
+<a name="reexecuting"/>
+
+## Reexecuting
```
niemeyer:
```
## Repeating runs

File: README.md [hunk 4]
```diff
@@ -516,6 +517,25 @@ as a `debug-each` script at the project, backend, and suite levels, so they
...
+
+Reproduce an error could be a very boring experience, and Spread has a way to
```
niemeyer: Reproducing an error may be

README.md
```diff
@@ -516,6 +517,25 @@ as a `debug-each` script at the project, backend, and suite levels, so they
...
+simplify that process by reexecuting the tasks as many times is desired until
```
niemeyer: s/is/as/

README.md
```diff
@@ -516,6 +517,25 @@ as a `debug-each` script at the project, backend, and suite levels, so they
...
+the task fails.
+
+To do that there are two options `-retries`, and `-stop-on-fail` which are
+used to determine the number of time to reexecute a task and if those executions
+have to be stoped when the task has failed.
+
+The `-retries` option receives an integer which indicates the number of retries 
+to do, being 0 the default value, and -1 can be used to reexecute indefinitely.
```
niemeyer: Let's please drop the -1 special behavior and that paragraph. People can easily type 100 or 10000.

README.md
```diff
@@ -516,6 +517,25 @@ as a `debug-each` script at the project, backend, and suite levels, so they
...
+
+All the tasks listed to be executed will be reexercuted following the same
```
niemeyer: That may be dropped as well. It's implied that they're supposed to run similarly.

spread/runner.go
```diff
@@ -594,27 +596,35 @@ func (r *Runner) worker(backend *Backend, system *System, order []int) {
...
-			r.add(&stats.TaskError, job)
-			debug = ""
-		}
-		if !abend && !r.options.Restore {
-			if err := r.fetchResidue(client, job); err != nil {
-				printf("Cannot fetch residue of %s: %v", job, err)
-				r.tomb.Killf("cannot fetch residue of %s: %v", job, err)
+		var retry = 0
+		var stopJob = false
+		for retry <= r.options.Retries && !stopJob {
```
niemeyer:
```
for repeat := r.options.Repeat; repeat >= 0; repeat-- {

File: spread/runner.go [hunk 9]
```diff
@@ -594,27 +596,35 @@ func (r *Runner) worker(backend *Backend, system *System, order []int) {
...
+			if r.options.Restore {
+				// Do not prepare or execute.
+			} else if !r.options.Restore && !r.run(client, job, preparing, job, job.Prepare(), debug, &abend) {
+				r.add(&stats.TaskPrepareError, job)
+				r.add(&stats.TaskAbort, job)
+				debug = ""
+				stopJob = r.options.StopOnFail
```
niemeyer:
```
repeat = -1
```
Same below.

README.md
niemeyer: Updated to "tasks" instead of "runs".

spread/runner.go
sergiocazzolato: Well in this case I need also to consider the scenario when I have to continue executing

niemeyer: Please see the earlier comment at the top. We don't want to repeat failing executions.

sergiocazzolato: ok, make sense

README.md
```diff
@@ -516,6 +517,18 @@ as a `debug-each` script at the project, backend, and suite levels, so they
 are aggregated and repeated for every task under them.
 
 
+<a name="reexecuting"/>
```
niemeyer: repeating

README.md
```diff
@@ -516,6 +517,18 @@ as a `debug-each` script at the project, backend, and suite levels, so they
...
+
+## Repeating runs
```
niemeyer: s/runs/tasks/ (sorry, I updated the comment but was probably too late)

cmd/spread/main.go
```diff
@@ -32,6 +32,7 @@ var (
 	discard     = flag.Bool("discard", false, "Discard reused servers without running")
 	residue     = flag.String("residue", "", "Where to store residual data from tasks")
 	seed        = flag.Int64("seed", 0, "Seed for job order permutation")
+	repeat      = flag.Int("repeat", 0, "The number of reexecutions for each task")
```
niemeyer: Per earlier comment:
```
repeat = flag.Int("repeat", 0, "Number of times to repeat each task")

File: spread/runner.go [hunk 13]
```diff
@@ -594,27 +595,32 @@ func (r *Runner) worker(backend *Backend, system *System, order []int) {
...
+				debug = ""
+				repeat = -1
+			} else if !r.options.Restore && r.run(client, job, executing, job, job.Task.Execute, debug, &abend) {
+				r.add(&stats.TaskDone, job)
+			} else if !r.options.Restore {
+				r.add(&stats.TaskError, job)
+				debug = ""
+				repeat = -1
+			}
+			if !abend && !r.options.Restore {
```
niemeyer:
```
... && repeat == 0
```
We don't want to fetch the same files every time.

spread/runner.go
```diff
@@ -594,27 +595,32 @@ func (r *Runner) worker(backend *Backend, system *System, order []int) {
...
-			r.add(&stats.TaskError, job)
-			debug = ""
-		}
-		if !abend && !r.options.Restore {
-			if err := r.fetchResidue(client, job); err != nil {
-				printf("Cannot fetch residue of %s: %v", job, err)
-				r.tomb.Killf("cannot fetch residue of %s: %v", job, err)
+		for repeat := r.options.Repeat; repeat >= 0; repeat-- {
+			if r.options.Restore {
+				// Do not prepare or execute.
```
niemeyer:
```
// Do not prepare or execute, and don't repeat.
repeat = -1
```
We might conflict the options at some point, but for now it at least doesn't make sense to repeat the restore phase alone.

PR #30, opened 2017-06-07
Snap as classic

rmescandon:
Added snap with classic confinement. Corrected Qemu paths to find the backend image, to fit snap or from sources installation, or even if provided other path using SPREAD_QEMU_PATH


rmescandon:
Another branch includes qemu backend in strict confinement

PR #31, opened 2017-06-16
Print the command output when a reboot fails

sergiocazzolato:
This change is done to add information about what happened in the
machine before a reboot in case the machine cannot be connected.

There are some errors like https://travis-ci.org/snapcore/snapd/builds/243225822#L3461 
which are very difficult to understand without any information about what happened before the
reboot. And for example in this situation is not possible to connect
anymore to this machine and/or get extra information after the reboot.


sergiocazzolato:
Here there is an example of the output:
https://paste.ubuntu.com/24874669/

PR #34, opened 2017-07-28
Add execution condition to suites and tasks

sergiocazzolato:
This change introduces the concept of condition, which is an script that
is evaluated before running a test or suite and based on its results the
task is executed or skipped.
The idea of this condition is to avoid adding "if" in the prepare-
execute-restore to exit.
A clear example where it can be used is the test main/refresh-all in
snapd project. In that case the condition should be:

condition: |
    [ "$TRUST_TEST_KEYS" = "true" ]

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,8 @@ Spread
 [Variants](#variants)  
 [Blacklisting and whitelisting](#blacklisting)  
 [Preparing and restoring](#preparing)  
-[Functions](#functions)
+[Conditions](#conditions)  
+[Functions](#functions)  
 [Rebooting](#rebooting)  
 [Timeouts](#timeouts)  
 [Fast iterations with reuse](#reuse)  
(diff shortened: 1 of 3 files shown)


sergiocazzolato:
@niemeyer I implemented this change to address the issue that we discussed about there are many exits in the script for the refresh tests.
Any suggestion is wellcome.

niemeyer:
Thanks for trying to sort out that issue, Sergio, but as we discussed online I don't think this helps much. This is just moving the if condition to a different (and much more constrained) field, with some peculiar syntax from a shell-script angle.

The solution to the issue we discussed is actually simpler, and it doesn't require any feature in spread: we just need to drop all the exit statements and reorganize the code to exit at the bottom of the function. This will improve the readability much.

sergiocazzolato:
@niemeyer, thanks for reviewing, gree with you this is not the best solution for that problem that we discussed, but this change also is fixing another problem that we have in the snapd tests, which is that we are adding if statements in the prepare, execute and restore in 18 tests to skip the execution in case the "TRUST_TEST_KEYS are false. 
This is a search with that example: https://paste.ubuntu.com/25233297/ 

A condition or something similar could help us to simplify those tests and others as well. The tests we discussed (refresh-*) are included on that list.

What do you think should be a good solution for those tests?

niemeyer:
Having a conditional statement would not prevent repeating the conditional all 18 times.

The traditional way to organize tests that have common needs is via suites. That prevents the duplication altogether.

Issue #38, opened 2017-10-24
Failed to allocate 17.* LXD systems


Saviq:
Given a backend:
```
    lxd:
        systems:
            - ubuntu-16.04
            - ubuntu-17.04
            - ubuntu-17.10
```

Spread fails to allocate the 17.* systems:
```
2017-10-24 15:18:39 Discarding lxd:ubuntu-17.04 (spread-12-ubuntu-17-04), cannot connect: cannot connect to lxd:ubuntu-17.04 (spread-12-ubuntu-17-04): ssh: handshake failed: ssh: unable to authenticate, attempted methods [none password], no supported methods remain
2017-10-24 15:18:43 Discarding lxd:ubuntu-17.10 (spread-14-ubuntu-17-10), cannot connect: cannot connect to lxd:ubuntu-17.10 (spread-14-ubuntu-17-10): ssh: handshake failed: ssh: unable to authenticate, attempted methods [none password], no supported methods remain
```

kyrofa:
This happens for 18.04 as well.

PR #39, opened 2017-10-24
Fix sshd_config sed expression (Issue #38)

Saviq:
The line might be commented out and have whitespace around.


niemeyer:
Thanks, and sorry for taking so long. I'll need to unify this at some point, but I'll just merge this for now.

Issue #11, opened 2016-08-26
spread (via snap) exec: "lxc": executable file not found in $PATH


ryan-beisner:
*Cannot allocate lxd:ubuntu-16.04: cannot launch lxd container: exec: "lxc": executable file not found in $PATH*

I was looking to start using spread very soon while we're redeploy some ci infra.  I've got a fresh Xenial instance.  I can launch and list lxc a-ok as my user, but can't get spread to do the examples.

It has the same error whether installed in strict or in dev mode.

```
ubuntu@osci-slavehost:~/temp⟫ tree
.
├── examples
│   └── hello
│       └── task.yaml
├── spread-hello.yaml
└── spread.yaml

2 directories, 3 files


ubuntu@osci-slavehost:~/temp⟫ head -n999 examples/hello/task.yaml spread-hello.yaml spread.yaml 
==> examples/hello/task.yaml <==
summary: Greet the planet
environment:
    GREETING: Hello
    GREETING/bar: Goodbye
    SUBJECT/baz: world
execute: |
    echo "$GREETING $SUBJECT!"
    exit 1

==> spread-hello.yaml <==
summary: Greet the planet
environment:
    SUBJECT: world 
    GREETING: Hello $SUBJECT!
execute: |
    echo "$GREETING"
    exit 1

==> spread.yaml <==
project: hello-world

backends:
    lxd:
        systems:
            - ubuntu-16.04

suites:
    examples/:
        summary: Simple examples
        environment:
            SUBJECT/foo: sanity
            SUBJECT/bar: lunacy

path: /remote/path


ubuntu@osci-slavehost:~/temp⟫ lxc list
+-----------------------+---------+------------------+------+------------+-----------+
|         NAME          |  STATE  |       IPV4       | IPV6 |    TYPE    | SNAPSHOTS |
+-----------------------+---------+------------------+------+------------+-----------+
| intergalactic-shellie | RUNNING | 10.0.8.45 (eth0) |      | PERSISTENT | 0         |
+-----------------------+---------+------------------+------+------------+-----------+

# ^ lxc unit created by ubuntu user manually


ubuntu@osci-slavehost:~/temp⟫ spread -list
lxd:ubuntu-16.04:examples/hello:baz
lxd:ubuntu-16.04:examples/hello:bar
lxd:ubuntu-16.04:examples/hello:foo
ubuntu@osci-slavehost:~/temp⟫ spread -v
[...]

PR #43, opened 2017-11-13
Fix LXD backend when run as a snap (Issue #11).

RAOF:
Snaps can't access binaries outside their shipped rootfs, so ship the lxc command
in the snap.

This requires that the user manually connect the lxd interface, but that's unavoidable.


mvo5:
I guess a generic "can we build a snap at all" would also be good.

PR #47, opened 2017-11-22
Add support for project-wide measure-each stanza

zyga:
The measurement feature allows to use a script to measure the testbed
(in some meaningful way) and raise a red flag (hopefully with a useful
error message) when something changes. This can be invaluable to look
for tests that don't clean up properly or change the system in some
unexpected way.

Signed-off-by: Zygmunt Krynicki <zygmunt.krynicki@canonical.com>


zyga:
This needs more work. I was working on this lately and I would like to close it for now.

PR #48, opened 2018-01-04
Fix Fedora images in LXD backend (Fixes #42, Fixes #44)

### LINKED ISSUE #42: Fails to start Fedora 26 LXD images, missing SSH
The Fedora 26 (at least) images on the LXD image servers do not come with openssh-server installed. This then causes spread to fail when it tries to mess with the ssh config and connect during image setup.

Assuming we don't want to hardcode the necessary ssh installation invocations in the lxd backend, maybe the lxd backend needs an extra setup step where the task can install the necessary packages?

### LINKED ISSUE #44: Fails to transfer project to Fedora image
The Fedora images on the LXD images: host don't ship with tar installed by default, resulting in a failure to execute `/bin/tar` when spread tries to send the project data across.

RAOF:


spread/lxd.go
```diff
@@ -67,6 +95,44 @@ func (s *lxdServer) Discard(ctx context.Context) error {
...
+
+	err = s.p.tuneSSH(s.d.Name, s.Distro())
+	if err != nil {
+		s.Discard(ctx)
+		return err
+	}
+
+	if s.Distro() == Fedora {
+		// Fedora LXD images do *not* contain tar by default, so we must install it manually
+		args = []string{"exec", s.d.Name, "--", "dnf", "install", "--assumeyes", "tar"}
```
Saviq: As discussed, this could be done by adding tar to the [fedora image setup script](https://github.com/lxc/lxc-templates/blob/master/templates/lxc-fedora.in#L610) instead.

spread/lxd.go
```diff
@@ -156,6 +221,17 @@ func isDebArch(s string) bool {
 	return false
 }
 
+func sshInstallCommand(distro Distro) []string {
+	if distro == Ubuntu || distro == Debian {
+		return []string{"apt", "install", "openssh-server"}
```
Saviq: This seems to be fixing Ubuntu and Debian, not Fedora, as the PR title suggests, but having spoken to people behind linux-containers.org, ssh servers are not installed in those images on purpose, for security reasons (default ssh server settings are often too open).

PR #53, opened 2018-03-18
Check uptime on system to determine if reboot was done

sergiocazzolato:
Compare uptime before and after reboot to determine if the reboot was done

### DIFF
--- a/spread/client.go
+++ b/spread/client.go
@@ -69,7 +69,7 @@ func (c *Client) ResetJob() {
 	c.SetJob("")
 }
 
-func (c *Client) dialOnReboot(prevUptime string) error {
+func (c *Client) dialOnReboot(prevUptime time.Time) error {
 	// First wait until SSH isn't working anymore.
 	timeout := time.After(c.killTimeout)
 	relog := time.NewTicker(c.warnTimeout)
@@ -79,6 +79,7 @@ func (c *Client) dialOnReboot(prevUptime string) error {
 
 	waitConfig := *c.config
 	waitConfig.Timeout = 5 * time.Second
+	uptimeDelta := 3 * time.Second
 
 	for {
 		before := time.Now()
@@ -90,8 +91,15 @@ func (c *Client) dialOnReboot(prevUptime string) error {
 
 		c.sshc.Close()
 		c.sshc = sshc
-		currUptime, _ := c.Output("uptime -s", "", nil)
-		if len(currUptime) > 0 && prevUptime != string(currUptime) {
+		currUptime, err := c.getUptime()
+		if err != nil {
+			// Not available uptime yet
+			continue
+		}
+
+		elapsedTime := currUptime.Sub(prevUptime)
+		if  elapsedTime > uptimeDelta {
+			// Reboot done
 			return nil
 		}
 		// Dial was observed not respecting the timeout by a long shot. Enforce it.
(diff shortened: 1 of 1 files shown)


spread/client.go
```diff
@@ -281,26 +281,32 @@ func (c *Client) run(script string, dir string, env *Environment, mode outputMod
 		rebootKey = rerr.Key
 		output = append(output, '\n')
 
-		timedout := time.After(c.killTimeout)
-		err := c.Run(fmt.Sprintf("reboot &\nsleep %.0f", c.killTimeout.Seconds()), "", nil)
+		uptime, err := c.getUptime()
 		if err != nil {
-			err = c.Run("echo should-have-disconnected", "", nil)
+			return nil, fmt.Errorf("failed to get uptime before reboot, error: %v", err)
```
niemeyer: This error message should be inside getUptime (with some tuning, will comment there).

spread/client.go
```diff
@@ -281,26 +281,32 @@ func (c *Client) run(script string, dir string, env *Environment, mode outputMod
...
-		if err == nil {
-			select {
-			case <-timedout:
-				return nil, fmt.Errorf("kill-timeout reached on %s while waiting for reboot", c.job)
-			default:
-			}
-			return nil, fmt.Errorf("reboot request on %s failed", c.job)
-		}
-		if err := c.dialOnReboot(); err != nil {
+		c.Run(fmt.Sprintf("reboot"), "", nil)
```
niemeyer: That's a pretty curious Sprintf. :)

spread/client.go
```diff
@@ -79,14 +79,29 @@ func (c *Client) dialOnReboot() error {
...
+
+		c.sshc.Close()
+		c.sshc = sshc
+		currUptime, err := c.getUptime()
+		if err != nil {
+			// Not available uptime yet
+			continue
+		}
+
+		elapsedTime := currUptime.Sub(prevUptime)
```
niemeyer: Elapsed makes it sound like the time since we started looking, but this is really the elapsed time between one boot and the other boot, so this is actually the real _uptimeDelta._ The test then becomes more readable:
```
if uptimeDelta > uptimeChanged { ... }
```

spread/client.go
```diff
@@ -79,14 +79,29 @@ func (c *Client) dialOnReboot() error {
 
 	waitConfig := *c.config
 	waitConfig.Timeout = 5 * time.Second
+	uptimeDelta := 3 * time.Second
```
niemeyer: Let's please make the use case for this variable more clear by naming it _uptimeChanged_.

spread/client.go
```diff
@@ -281,26 +281,32 @@ func (c *Client) run(script string, dir string, env *Environment, mode outputMod
...
-			select {
-			case <-timedout:
-				return nil, fmt.Errorf("kill-timeout reached on %s while waiting for reboot", c.job)
-			default:
-			}
-			return nil, fmt.Errorf("reboot request on %s failed", c.job)
-		}
-		if err := c.dialOnReboot(); err != nil {
+		c.Run(fmt.Sprintf("reboot"), "", nil)
+		if err := c.dialOnReboot(uptime); err != nil {
```
niemeyer: Let's please have a `time.Sleep(3 * time.Second)` before the line above. Dialing right after running the reboot command will add more churn to the rebooting system for no benefit.

spread/client.go
```diff
@@ -281,26 +281,32 @@ func (c *Client) run(script string, dir string, env *Environment, mode outputMod
...
 }
 
+func (c *Client) getUptime() (time.Time, error) {
+	uptime, err := c.Output("date -u --iso-8601=seconds -d \"`cut -f1 -d. /proc/uptime` seconds ago\"", "", nil)
+	if err != nil {
+		return time.Now(), err
+	}
+
+	// Convert unsupported +0000 to +00:00
+	parts := strings.Split(string(uptime), "+")
```
niemeyer: We can pick the format so we don't need to do this:
```
date -u -d "$(cut -f1 -d. /proc/uptime) seconds ago" +"%Y-%m-%dT%H:%M:%SZ"
```

spread/client.go
```diff
@@ -281,26 +281,32 @@ func (c *Client) run(script string, dir string, env *Environment, mode outputMod
...
 			return nil, err
 		}
 	}
 	panic("unreachable")
 }
 
+func (c *Client) getUptime() (time.Time, error) {
+	uptime, err := c.Output("date -u --iso-8601=seconds -d \"`cut -f1 -d. /proc/uptime` seconds ago\"", "", nil)
+	if err != nil {
+		return time.Now(), err
```
niemeyer: `time.Time{}` is the zero value of `time.Time`.

Also, let's please return an actual error message, both here and below, if we cannot get the time:
```
return time.Time{}, fmt.Errorf("cannot obtain the remote system uptime: %v", err)
```

spread/client.go
```diff
@@ -281,26 +281,32 @@ func (c *Client) run(script string, dir string, env *Environment, mode outputMod
...
+	}
+
+	// Convert unsupported +0000 to +00:00
+	parts := strings.Split(string(uptime), "+")
+	if (len(parts) == 2) && (! strings.Contains(parts[1], ":")) {
+		parts[1] = "00:00"
+	}
+	return time.Parse(time.RFC3339, strings.Join(parts, "+"))
```
niemeyer:
```
if err != nil {
    return time.Time{}, fmt.Errorf("cannot parse the remote system uptime: %q", uptime)
```
Note that in this case the value after the colon is the actual time, not the error. That's what we need to tell what went wrong.

spread/client.go
```diff
@@ -281,26 +281,34 @@ func (c *Client) run(script string, dir string, env *Environment, mode outputMod
...
 
+func (c *Client) getUptime() (time.Time, error) {
+	uptime, err := c.Output("date -u -d \"$(cut -f1 -d. /proc/uptime) seconds ago\" +\"%Y-%m-%dT%H:%M:%SZ\"", "", nil)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("cannot obtain the remote system uptime: %v", err)
+	}
+
+	parsedUptime, err := time.Parse(time.RFC3339, string(uptime))
+	if err != nil {
+		return time.Time{}, fmt.Errorf("cannot parse the remote system uptime: %q", parsedUptime)
```
niemeyer: On error the parsed uptime will be a zero time.

PR #55, opened 2018-03-22
Make possible to add an initialization script for custom images

sergiocazzolato:
This will help to customiza the initialization for custom images.


niemeyer:
A few different points about this:

1. The initialization script in Spread is the content of the "prepare:" field.

2. If you are baking an image in Google, why would we put content on a spread-specific location instead of having it in the usual place.

3. Given the context that we've been discussing, it's still not clear to me why we need this. You said earlier today you had solved the problem?

Let's please talk tomorrow about this.

PR #57, opened 2018-04-14
Adding preserve option for systems to keep the default image size

sergiocazzolato:
In case preserve option is true, then the system image will keep the default size, otherwise it will be set to 10 GB by default.

### DIFF
--- a/spread/google.go
+++ b/spread/google.go
@@ -428,6 +428,18 @@ func (p *googleProvider) createMachine(ctx context.Context, system *System) (*go
 		})
 	}
 
+	initializeParams := googleParams{}
+	if system.Preserve == true {
+		initializeParams = googleParams{
+			"sourceImage": sourceImage,
+		}
+	} else {
+		initializeParams = googleParams{
+			"sourceImage": sourceImage,
+			"diskSizeGb": storage,
+		}
+	}
+
 	params := googleParams{
 		"name":        name,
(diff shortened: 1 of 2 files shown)


bboozzoo:
@niemeyer would it be possible for you to review this soon(-ish)? This is required for moving forward with Amazon Linux 2 testing.

spread/google.go
```diff
@@ -428,6 +428,18 @@ func (p *googleProvider) createMachine(ctx context.Context, system *System) (*go
 		})
 	}
 
+	initializeParams := googleParams{}
+	if system.Preserve == true {
```
niemeyer: The idea of "preserve" was to be a storage size setting, as in
```
storage: preserve
```
The above literally reads as "preserve the system?", which is unrelated to the feature.

In fact, we can do even better than that and have
```
storage: image-size
```
Which is much more clear in intention.

As an implementation note, the _storage_ variable in this block of code is solely responsible for tuning the disk size in this map below. It's awkward to have it up there in the code, and then down here we again consider what to do about that exact value.

I've fixed those issues in the commit https://github.com/snapcore/spread/commit/070a7c4. Please let me know if that works well for the intended purpose.

bboozzoo: Thanks for fixing it!

niemeyer: Please note I've changed from "image-size" to "preserve-size" above, as I realized that the Size type is not specific to this context, and it wouldn't make sense to have "image-size" elsewhere.

PR #62, opened 2018-07-02
client: fix dialOnReboot and add (regression) test

mvo5:
Based on #61 - not super clean but I can rework to improve it - with the old code this test will timeout.


spread/client_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/snapcore/spread/spread"
+)
+
+type clientSuite struct{}
+
+var _ = Suite(&clientSuite{})
+
+func (s *clientSuite) TestDialOnReboot(c *C) {
+	restore := spread.MockSshDial(func(network, addr string, config *ssh.ClientConfig) (*ssh.Client, error) {
+		time.Sleep(1 * time.Second)
```
zyga: Initially I reacted with "let's make this a channel and remove the sleep" but reading the rest made me realize it's not so obvious. I think this is okay, until we figure out a way to write such tests without any sleeps at all.

I was thinking briefly about something like this:
```
spread.MockWarnCondition(...) // replaces internal sleep with receive from channel.
spread.MockKillCondition(...) // replaces internal sleep with receive from another channel
```

Then you could really receive from a channel in the `MockSshDial` and that would be race-free and fast.

PR #64, opened 2018-07-20
 spread: disable -o pipefail when doing the echo|grep

mvo5:
With SIGPIPE and "echo | grep -q" there is a race condition. If
grep exists before echo is finished writing data then echo will
get a SIGPIPE and this makes MATCH fail (which is very obscure
to the user).

This PR fixes it by disabling pipefail for MATCH.

It also "unrolls" the MATCH helper because it gets a bit hard to read in the
"inline" form that was used before.

### DIFF
--- a/spread/client.go
+++ b/spread/client.go
@@ -361,7 +361,7 @@ func (c *Client) runPart(script string, dir string, env *Environment, mode outpu
 	}
 	buf.WriteString(rc(false, "REBOOT() { { set +xu; } 2> /dev/null; [ -z \"$1\" ] && echo '<REBOOT>' || echo \"<REBOOT $1>\"; exit 213; }\n"))
 	buf.WriteString(rc(false, "ERROR() { { set +xu; } 2> /dev/null; [ -z \"$1\" ] && echo '<ERROR>' || echo \"<ERROR $@>\"; exit 213; }\n"))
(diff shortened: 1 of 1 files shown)


spread/client.go
```diff
@@ -22,6 +22,32 @@ import (
...
+
+    local stdin="$(cat)"
+
+    # disable pipefail temporarily, "echo" can get SIGPIPE in the pattern below
+    pf="$(shopt -po pipefail)"
+    set +o pipefail
+    # do the grep
+    echo "$stdin" | grep -q -E "$@" || res=$?
+    # restore pipefail
+    eval "$pf"
```
chipaca: we could instead do
```sh
grep -q -E "$@" <<<"$stdin" || res=$?
```
with no pipes

spread/client.go
```diff
@@ -22,6 +22,27 @@ import (
...
+var match string = `MATCH() {
+    {
+        set +xu
+    } 2> /dev/null
+
+    [ ${#@} -gt 0 ] || { echo "error: missing regexp argument"; return 1; }
+
+    local stdin="$(cat)"
+    # We are not using pipes here, see:
+    #   https://github.com/snapcore/spread/pull/64
```
niemeyer: Can we keep this comment in the Go code and out of the actual string?

It'd also be nice to have just the actual fix in place for this PR, so that we can clearly see the fix independent from the change. I realize the logic is a bit cryptic, but that's done in part because this string is sent along with every single small script that is run on the server. It's usually not the focus.

PR #65, opened 2018-07-25
spread: simplify MATCH by using "<<<" (thanks to chipaca)

mvo5:
With SIGPIPE and "echo | grep -q" there is a race condition. If
grep exists before echo is finished writing data then echo will
get a SIGPIPE and this makes MATCH fail (which is very obscure
to the user). To fix this this PR switch to the "<<<" HERE
string.

### DIFF
--- a/spread/client.go
+++ b/spread/client.go
@@ -361,7 +361,7 @@ func (c *Client) runPart(script string, dir string, env *Environment, mode outpu
 	}
 	buf.WriteString(rc(false, "REBOOT() { { set +xu; } 2> /dev/null; [ -z \"$1\" ] && echo '<REBOOT>' || echo \"<REBOOT $1>\"; exit 213; }\n"))
 	buf.WriteString(rc(false, "ERROR() { { set +xu; } 2> /dev/null; [ -z \"$1\" ] && echo '<ERROR>' || echo \"<ERROR $@>\"; exit 213; }\n"))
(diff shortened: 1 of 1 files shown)


spread/client.go
```diff
@@ -361,7 +361,9 @@ func (c *Client) runPart(script string, dir string, env *Environment, mode outpu
 	}
 	buf.WriteString(rc(false, "REBOOT() { { set +xu; } 2> /dev/null; [ -z \"$1\" ] && echo '<REBOOT>' || echo \"<REBOOT $1>\"; exit 213; }\n"))
 	buf.WriteString(rc(false, "ERROR() { { set +xu; } 2> /dev/null; [ -z \"$1\" ] && echo '<ERROR>' || echo \"<ERROR $@>\"; exit 213; }\n"))
-	buf.WriteString(rc(true, "MATCH() { { set +xu; } 2> /dev/null; [ ${#@} -gt 0 ] || { echo \"error: missing regexp argument\"; return 1; }; local stdin=\"$(cat)\"; echo \"$stdin\" | grep -q -E \"$@\" || { echo \"error: pattern not found, got:\n$stdin\">&2; return 1; }; }\n"))
+	// We are not using pipes here, see:
+	//  https://github.com/snapcore/spread/pull/64
+	buf.WriteString(rc(true, "MATCH() { { set +xu; } 2> /dev/null; [ ${#@} -gt 0 ] || { echo \"error: missing regexp argument\"; return 1; }; local stdin=\"$(cat)\"; grep -q -E \"$@\" <<< \"$stdin\"  || { res=$?; echo \"error: pattern not found, got:\n$stdin\">&2; if [ $res != 1 ]; then echo \"unexpected exit status: $res\"; fi; return 1; }; }\n"))
```
mvo5: Feedback from Gustavo: use: `grep error: %d`  and fix the extra spaces

PR #66, opened 2018-07-31
Define storage at system level

sergiocazzolato:
This is to be able to set the preserve-size option for each storage independently.

Currently for snpad test suite we are using "storage: preserve-size" at backend level but it is forcing all the systems to preserve the storage size even when it is not needed, which could produce errors testing images with are 1 or 3 GB of storage such as some imported images from other projects. 

This is an example of the error that this is causing: https://paste.ubuntu.com/p/HJcSgvBp4b/
It is happening installing tests dependencies on ubuntu-16.04-32 on spread-images project.

### DIFF
--- a/spread/project.go
+++ b/spread/project.go
@@ -544,6 +544,9 @@ func Load(path string) (*Project, error) {
 			if system.Workers == 0 {
 				system.Workers = 1
 			}
+			if system.Storage == 0 {
+				system.Storage = backend.Storage
+			}
 			if err := checkEnv(system, &system.Environment); err != nil {
 				return nil, err
 			}
--- a/spread/project.go
+++ b/spread/project.go
@@ -119,6 +119,9 @@ type System struct {
 	Password string
 	Workers  int
 
+	// Only for Linode and Google so far.
+	Storage  Size
+
 	Environment *Environment
 	Variants    []string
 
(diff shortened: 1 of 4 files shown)


spread/project.go
```diff
@@ -119,6 +119,9 @@ type System struct {
 	Password string
 	Workers  int
 
+	// Only for Linode and Google so far.
+	Storage  Size
```
niemeyer: This setting should cascade from the backend when the backend has it set, so that the actual backend implementations don't need to check both. There are good examples in how other options are handled (Priority, Manual, etc).

sergiocazzolato: @niemeyer thanks for reviewing this, I already implemented the change.

spread/runner.go
```diff
@@ -881,6 +881,10 @@ func (r *Runner) allocateServer(backend *Backend, system *System) *Client {
 Allocate:
 	for {
 		lerr := err
+
+		if system.Storage == 0 {
+			system.Storage = backend.Storage
+		}
```
niemeyer: Right, that's the idea, but not here. When loading the project:

https://github.com/snapcore/spread/blob/52e4138/spread/project.go#L546

Also, as a general note every time we add something to the code we need to be conscious about what is around the change. The change above broke a line out from the block which it was related to (lerr := ...) and inserted something in between which was completely unrelated.

sergiocazzolato: @niemeyer right, thanks for taking a look to this. I already made the change.

PR #67, opened 2018-08-02
spread: run MATCH in a subshell

mvo5:
We ran into a confusing error when using MATCH in one of our snapd
tests. It runs out the issue can be reproduced with:
```
MATCH () {
    {
        set +ux
    } 2> /dev/null;
    return 0
}
MATCH foo <<< foo
test 1 -eq 2
```
When this is used, the output is:
```
$ bash -eux reproducer.sh
+ MATCH foo
```
I.e. the test line is not displayed.

It turns out the reason is that the MATCH function uses "set +ux"
which turns off the traces globally. When MATCH is used in a
pipe (e.g. "echo foo | MATCH foo") then MATCH is run in a
subshell and the turning off is not criticial. However when
using with here strings for file redirects and no subshell
is used turning off trace globally is an issue. The fix is
simply to run MATCH always in a subshell.

### DIFF
--- a/spread/client.go
+++ b/spread/client.go
@@ -363,7 +363,9 @@ func (c *Client) runPart(script string, dir string, env *Environment, mode outpu
 	buf.WriteString(rc(false, "ERROR() { { set +xu; } 2> /dev/null; [ -z \"$1\" ] && echo '<ERROR>' || echo \"<ERROR $@>\"; exit 213; }\n"))
 	// We are not using pipes here, see:
 	//  https://github.com/snapcore/spread/pull/64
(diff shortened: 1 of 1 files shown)


sergiocazzolato:
@mvo5 I could reproduce the error using this PR :(

mvo5:
@sergiocazzolato Do you have more details? Maybe it is a different error or a different incarnation of the errror?

sergiocazzolato:
@mvo5 this is the log I have: https://paste.ubuntu.com/p/Xxg5qVc6m5/ It seems to be the same than we see on travis

sergiocazzolato:
@mvo5 I tested this fix on core18 with snap-env test which was failing doing MATCH 'foo' < file, and works very well.

niemeyer:
Very nice find, thanks Michael!

PR #69, opened 2018-09-18
Manage pagination retrieving images from google backend

sergiocazzolato:
Currently we are failing to get the fedora base image because it is in the second page of images retrieved by google. 

See spread output: 

spread google:fedora-28-64-base:tasks/google/update-fedora-28-64
2018-09-17 06:54:44 Project content is packed for delivery (112.64KB).
2018-09-17 06:54:44 If killed, discard servers with: spread -reuse-pid=27395 -discard
2018-09-17 06:54:44 Allocating google:fedora-28-64-base...
2018-09-17 06:54:44 Cannot allocate google:fedora-28-64-base: cannot find any Google image matching "fedora-28-64-base" on project "computeengine"
2018-09-17 06:54:44 Successful tasks: 0
2018-09-17 06:54:44 Aborted tasks: 1
error: unsuccessful run

### DIFF
--- a/spread/google.go
+++ b/spread/google.go
@@ -31,6 +31,8 @@ func Google(p *Project, b *Backend, o *Options) Provider {
 		options: o,
 
 		imagesCache: make(map[string]*googleImagesCache),
+
+		apiURL: "https://www.googleapis.com",
 	}
 }
 
@@ -50,6 +52,9 @@ type googleProvider struct {
 	keyErr     error
 
 	imagesCache map[string]*googleImagesCache
+
+	// allow mocking in tests
+	apiURL string
 }
 
 type googleServer struct {
@@ -336,12 +341,12 @@ type googleImagesCache struct {
 
 type googleImageListResult struct {
 	Items []struct {
-		Description	string
-		Status		string
-		Name		string
-		Family		string
+		Description string
+		Status      string
+		Name        string
+		Family      string
 	}
-	NextPageToken	string
+	NextPageToken string
 }
 
 func (p *googleProvider) projectImages(project string) ([]googleImage, error) {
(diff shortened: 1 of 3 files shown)


spread/google_test.go
```diff
@@ -0,0 +1,78 @@
...
+	defer mockServer.Close()
+
+	g := spread.NewGoogleProviderForTesting(mockServer.URL, nil, nil, nil)
+	c.Assert(g, NotNil)
+
+	images, err := g.ProjectImages("snapd")
+	c.Assert(err, IsNil)
+	c.Check(images, HasLen, 5)
+	i := 0
+	c.Check(images[i].Project, Equals, "snapd")
```
anonymouse64: nitpick: personally I think this would read better as a DeepEquals with the full images object compared, but this is okay too

spread/google.go
```diff
@@ -350,32 +366,39 @@ func (p *googleProvider) projectImages(project string) ([]googleImage, error) {
...
+			"orderBy":   {"creationTimestamp+desc"},
+			"pageToken": {nextPageToken},
 		}
-	}
+		err := p.dofl("GET", "/compute/v1/projects/"+project+"/global/images?"+v.Encode(), nil, &result, noPathPrefix)
+		nextPageToken = result.NextPageToken
+		if err != nil {
+			return nil, &FatalError{fmt.Errorf("cannot retrieve Google images for project %q: %v", project, err)}
+		}
+		results = append(results, &result)
```
anonymouse64: nitpick: I think this would read cleaner (and would eliminate the double for loop below if we had `results` as a single `googleImageListResult` and here did

```suggestion
		results.Items = append(results.Items, &result.Items...)
```
I don't think we need to save the `nextPageToken` at all for outside the for loop, so this simplifies storing the results inside a single list rather than a list of structures with lists inside the structures

anonymouse64: also if you do take this, probably rename `results` -> `finalResults` so it reads a bit better:
```suggestion
		finalResults.Items = append(finalResults.Items, result.Items...)
```

spread/google_test.go
mvo5: Totally agreed. This is a bit tricky because the struct is "googleImage" which is not exported. As this is a "spread_test" style test (i.e. no access to internals of the spread package) I couldn't do a DeepEqual because I can't do:
```
c.Check(images, DeepEquals, spread.googleImage{...})
```
Putting `type GoogleImage googleImage` into export_test.go also does not help. I could simply export the type but not sure if that is to Gustavos liking so I kept things in there as much as possible :)

anonymouse64: Ah right I didn't realize that, in that case this looks good, thanks for adding a comment about it

mvo5:
Feedback from irc:
<niemeyer> mvo: Isn't the pagination design/API a general feature of GCE?
<niemeyer> mvo: In other words, will we be doing the same for every similar call?
<niemeyer> mvo: If so, should we integrate the feature in the 'do' method instead?
<niemeyer> mvo: I appreciate that in the original design the fiddling with the document is abstracted away: give a URL, obtain a Go value
<niemeyer> mvo: This is moving that slightly up the chain

sergiocazzolato:
It is implemented by #87

PR #70, opened 2018-11-05
Reboot on backgraound to avoid spread wait for long time

sergiocazzolato:
This is to avoid this messages showing the output still empty when the
reboot was already done:

2018-11-05 14:48:35 Rebooting on nov051744-506863
(google:debian-9-64-base:tasks/google/update-debian-9) as requested...
2018-11-05 14:48:35 Getting uptime
2018-11-05 14:48:35 Uptime calculated
2018-11-05 14:48:35 Running reboot command
2018-11-05 14:53:36 WARNING: nov051744-506863
(google:debian-9-64-base:tasks/google/update-debian-9) running late.
Output still empty.
2018-11-05 14:58:36 WARNING: nov051744-506863
(google:debian-9-64-base:tasks/google/update-debian-9) running late.
Output still empty.

It can be reproduced easy on debian-9 where the reboot is done very fast
(~8sec).

With this chage the reboot is done on backgrand, and the ssh always
return. Then the uptime is compared to make sure the rebook was properly
done.

### DIFF
--- a/spread/client.go
+++ b/spread/client.go
@@ -296,7 +296,7 @@ func (c *Client) run(script string, dir string, env *Environment, mode outputMod
 		if err != nil {
 			return nil, err
 		}
-		c.Run("reboot", "", nil)
+		c.Run("reboot -f >/dev/null &", "", nil)
 
 		if err := c.dialOnReboot(uptime); err != nil {
 			return nil, err


sergiocazzolato:
@niemeyer could you please take a look to this one?

Thanks

sergiocazzolato:
@niemeyer, I already added the change to detect when the reboot command is stuck, and in that case we manually close the ssh connection. 
I already validated this with the images debian-9-stretch and debian-9-64-base.

sergiocazzolato:
This is the new log: https://paste.ubuntu.com/p/YgY3HJ8b7d/

sergiocazzolato:
This is an error that heppened today: https://paste.ubuntu.com/p/kzJhnP7BHj/

spread/client.go
```diff
@@ -296,7 +296,18 @@ func (c *Client) run(script string, dir string, env *Environment, mode outputMod
...
+		go func() {
+			errch <- c.Run("reboot", "", nil)
+		}()
+
+		warn := time.NewTicker(c.warnTimeout)
+		select {
+		case <-errch:
+		case <-warn.C:
+			c.sshc.Close()
+		}
```
chipaca: here you're leaving the ticker running. This is probably not a huge problem, but it would be tidiest to do a
```go
warn.Stop()
```
in the `errch` case. On the other hand, if you don't think that's needed, you could've instead done directly
```go
case <- time.Tick(c.warnTimeout)
```

PR #75, opened 2019-03-14
Make spread tests for spread project run on google backend

sergiocazzolato:
This change how the project is prepared and introduce some tests fixes
and a fix in the code as well to make the residue test work again.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -46,8 +46,7 @@ suites:
 
 prepare: |
     apt update
-    apt install -y git qemu-kvm snapd
-    snap install lxd tree
+    apt install -y git qemu-kvm lxd tree
 
     # Cache is only available if SEND_CACHE=1 is set.
     mkdir -p tests/cache
--- a/spread.yaml
+++ b/spread.yaml
@@ -26,9 +26,6 @@ backends:
 
     qemu:
         systems:
-            - ubuntu-16.04-64:
-                username: ubuntu
-                password: ubuntu
             - ubuntu-18.04-64:
                 username: ubuntu
                 password: ubuntu
@@ -45,8 +42,12 @@ suites:
         summary: Integration tests
 
 prepare: |
-    apt update
-    apt install -y git qemu-kvm lxd tree
+
+    DEBS="git qemu-kvm lxd tree"
+    if ! dpkg -l $DEBS; then
+        apt update
+        apt install -y $DEBS
+    fi
 
     # Cache is only available if SEND_CACHE=1 is set.
     mkdir -p tests/cache
@@ -68,10 +69,10 @@ prepare: |
     go get ./cmd/spread
 
     # Cache is only available if SEND_CACHE=1 is set, follow the steps:
-    # . Create dest dir: > mkdir tests/cache/lxd-ubuntu-16.04
-    # . Get image id: > lxc image list
-    # . Export image: > lxc image export <IAMGE-ID> tests/cache/lxd-ubuntu-16.04
-    # . Extract matadata: > cd tests/cache/lxd-ubuntu-16.04 && tar -xvf <MATADATA-FILE>
+    # - Create dest dir: > mkdir tests/cache/lxd-ubuntu-16.04
+    # - Get image id: > lxc image list
+    # - Export image: > lxc image export <IAMGE-ID> tests/cache/lxd-ubuntu-16.04
+    # - Extract metadata: > cd tests/cache/lxd-ubuntu-16.04 && tar -xvf <METADATA-FILE>
     if [ -d tests/cache/lxd-ubuntu-16.04 ]; then
         if ! lxc image list | grep '16\.04'; then
             lxc image import tests/cache/lxd-ubuntu-16.04
@@ -80,6 +81,5 @@ prepare: |
 
     # Start lxd instance
     lxd init --auto
-    lxc launch ubuntu:16.04
 
 # vim:ts=4:sw=4:et
--- a/spread.yaml
+++ b/spread.yaml
@@ -26,6 +26,9 @@ backends:
 
     qemu:
         systems:
+            - ubuntu-16.04-64:
+                username: ubuntu
+                password: ubuntu
             - ubuntu-18.04-64:
                 username: ubuntu
                 password: ubuntu
@@ -43,15 +46,11 @@ suites:
 
 prepare: |
     apt update
-    apt install -y git lxd qemu-kvm tree
+    apt install -y git lxd tree qemu-kvm
 
     # Cache is only available if SEND_CACHE=1 is set.
     mkdir -p tests/cache
-    
-    # Start lxd instance
-    lxd init --auto
-    lxc launch ubuntu:16.04
-    
+
     # Cache Go deb with: cd tests/cache && godeb download $GOVERSION
     if ! dpkg -l go; then
         if [ ! -f tests/cache/go_$GOVERSION-*.deb ]; then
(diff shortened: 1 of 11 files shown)


sergiocazzolato:
@mvo5 @niemeyer Please take a look to the change.
To run the tests just do: spread google:ubuntu-18.04-64:
Thanks

tests/qemu/task.yaml
```diff
@@ -6,8 +6,11 @@ prepare: |
         exit 1
     fi
 
+    apt install -y autopkgtest genisoimage
+    autopkgtest-buildvm-ubuntu-cloud -a amd64 -r xenial
+
     mkdir -p ~/.spread/qemu
-    ln -sf /dev/sda ~/.spread/qemu/ubuntu-16.04.img
+    mv autopkgtest-xenial-amd64.img ~/.spread/qemu/ubuntu-16.04.img
```
niemeyer: What we used to have here would use a cached image. This is building the image every single time we run this test. If we have to debug this, we'll be paying the price of every single time too.

spread/runner.go
```diff
@@ -631,7 +631,7 @@ func (r *Runner) worker(backend *Backend, system *System, order []int) {
 				debug = ""
 				repeat = -1
 			}
-			if !abend && !r.options.Restore && repeat == 0 {
+			if !abend && !r.options.Restore && repeat <= 0 {
```
niemeyer: It's a very good practice to review one's own suggested change before others do.

spread.yaml
```diff
@@ -61,19 +68,4 @@ prepare: |
...
-    LXD_IPV4_DHCP_MAX="250"
-    LXD_IPV4_NAT="true"
-    END
-        systemctl stop lxd lxd-bridge
-        systemctl start lxd lxd-bridge
-    fi
-
-
-# vim:ts=4:sw=4:et
+# vim:ts=4:sw=4:et
```
niemeyer: Missing EOL.

spread.yaml
```diff
@@ -32,19 +42,16 @@ suites:
...
-    # Cache image with: mkdir tests/cache/lxd-ubuntu-16.04 && lxc export ubuntu:16.04 tests/cache/lxd-ubuntu-16.04
-    if ! lxc image list | grep '16\.04'; then
-        lxc image import tests/cache/lxd-ubuntu-16.04/meta-* tests/cache/lxd-ubuntu-16.04/[a-f0-9]* os=ubuntu version=16.04 architecture=amd64 remote=ubuntu
-    fi
-
+    mkdir -p tests/cache
+    
+    # Start lxd instance
+    lxd init --auto
+    lxc launch ubuntu:16.04
```
niemeyer: The original logic was building a cache and allowing one to use an existing cache. The new one has blown that away. It would be good to preserve the ability to use a cached image, and test it to make sure it still works.  Otherwise every time we run tests with qemu, we'll be paying the price of redownloading an image.

spread.yaml
```diff
@@ -8,15 +8,25 @@
...
+    google:
+        key: "$(HOST: echo $SPREAD_GOOGLE_KEY)"
+        location: computeengine/us-east1-b
+        halt-timeout: 2h
+        systems:
+            - ubuntu-18.04-64:
+                image: ubuntu-1804-64-virt-enabled
+                workers: 1
+
     qemu:
```
niemeyer: Is this backend still working?

tests/envs/task.yaml
```diff
@@ -13,3 +13,6 @@ execute: |
 
 debug: |
     cat task.out || true
+
+restore: |
+    rm -f .spread-reuse.yaml /run/spread-reuse.yaml
```
niemeyer: That doesn't sound right.  What's the point of using --reuse if it's blowing the data that allows reuse away?   Why is --reuse there in the first place?

tests/qemu/task.yaml
sergiocazzolato: Ok, I was thinking on run this just on google, but we really need also to be able to run that on qemu, I'll add the logic to keep both backends working.
Thanks for reviewing this

spread/runner.go
sergiocazzolato: This change is to make residue work when the test fails, currently it was just leaving the residue when the test passed. I noticed that when I ran the residue test and it was failing.

spread.yaml
sergiocazzolato: Sure, I'll restore the cache.
Thanks

tests/envs/task.yaml
sergiocazzolato: Agree, I need just to delete .spread-reuse.yaml to be able to run with -repeat option.

spread.yaml
sergiocazzolato: It is not working for me on bionic with qemu 2.11.1, bu works using qemu on xenial.

spread/runner.go
niemeyer: This seems unrelated to this PR purpose.

spread.yaml
niemeyer: Does it still work?

spread.yaml
niemeyer: Missing response here.

sergiocazzolato: sorry, qemu still works

spread/runner.go
sergiocazzolato: @niemeyer, Without this change the residue test is going to fail on google and qemu as well. I dont see any problem to revert that and add it in another PR. What do you prefer?

niemeyer: Ack, thanks for the explanation online.

spread.yaml
sergiocazzolato: @niemeyer  Yes, it is working

sergiocazzolato:
@niemeyer, hi, the PR is ready. Thanks

spread.yaml
```diff
@@ -32,18 +45,11 @@ suites:
         summary: Integration tests
 
 prepare: |
-    DEBS="git lxd kvm"
-    if ! dpkg -l $DEBS; then
-        apt-get update
-        apt-get install -y $DEBS
-    fi
+    apt update
+    apt install -y git qemu-kvm lxd tree
```
niemeyer: The original logic here would prevent the time consuming apt-get update/install every time if the debs were already installed. Was there a reason for the modification?

spread.yaml
```diff
@@ -61,19 +67,19 @@ prepare: |
...
-    LXD_IPV4_DHCP_MAX="250"
-    LXD_IPV4_NAT="true"
-    END
-        systemctl stop lxd lxd-bridge
-        systemctl start lxd lxd-bridge
+    # Cache is only available if SEND_CACHE=1 is set, follow the steps:
+    # . Create dest dir: > mkdir tests/cache/lxd-ubuntu-16.04
+    # . Get image id: > lxc image list
+    # . Export image: > lxc image export <IAMGE-ID> tests/cache/lxd-ubuntu-16.04
+    # . Extract matadata: > cd tests/cache/lxd-ubuntu-16.04 && tar -xvf <MATADATA-FILE>
```
niemeyer: s/mata/meta/

Can we please also have "-" as the list item prefix instead of "."

Also, why that tar in the last command?  Seems surprising that `lxc image import` cannot import what `lxc image export` exports.

spread.yaml
```diff
@@ -8,17 +8,30 @@
...
+            - ubuntu-18.04-64:
+                image: ubuntu-1804-64-virt-enabled
+                workers: 1
+
     qemu:
         systems:
             - ubuntu-16.04-64:
                 username: ubuntu
                 password: ubuntu
+            - ubuntu-18.04-64:
```
niemeyer: Are both systems still necessary? If you tested with the latter, we can probably drop the former.

spread.yaml
```diff
@@ -61,19 +67,19 @@ prepare: |
...
+    # . Extract matadata: > cd tests/cache/lxd-ubuntu-16.04 && tar -xvf <MATADATA-FILE>
+    if [ -d tests/cache/lxd-ubuntu-16.04 ]; then
+        if ! lxc image list | grep '16\.04'; then
+            lxc image import tests/cache/lxd-ubuntu-16.04
+        fi
     fi
 
+    # Start lxd instance
+    lxd init --auto
+    lxc launch ubuntu:16.04
```
niemeyer: Why is this required now?

spread.yaml
sergiocazzolato: No, I just changed that to get updates. I'll revert that.

spread.yaml
sergiocazzolato: Agree, it is super weird but this is a lxd known issue and I found the solution in a forum.

spread.yaml
sergiocazzolato: No, I'll leave 18.04, thanks

spread.yaml
sergiocazzolato: sorry, is is not required.

sergiocazzolato:
@niemeyer comments addressed.

Thanks for the review

mvo5:
Fwiw, I just ran the tests on qemu and GCE and everything is fine: https://paste.ubuntu.com/p/xR5dWRrmbW/ and

PR #79, opened 2019-06-12
New -workers option used to define the number of workers by system

sergiocazzolato:
The number of workers can be set as part of the spread.yaml file and
also by using the `-workers` option, which receives an integer
indicating the number of workers used by all the systems selected for
the run.

In case the `-workers` option is not used, the number of workers used is
the provided for the system in the spread.yaml which is 1 by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Spread
 [Fast iterations with reuse](#reuse)  
 [Debugging](#debugging)  
 [Repeating tasks](#repeating)
+[Define number of Workers](#workers)
 [Passwords and usernames](#passwords)  
 [Including, excluding, and renaming files](#including)  
 [Selecting which tasks to run](#selecting)  
@@ -560,6 +561,16 @@ the task fails.
 To do that there is an option `-repeat` which receives an integer indicating 
 the number of reexecutions to do, being 0 the default value.
 
+<a name="workers"/>
+
+## Define number of Workers
+
+The number of workers can be set as part of the spread.yaml file and also by
+using the `-workers` option, which receives an integer indicating
+the number of workers used by all the systems selected for the run.
+
+In case the `-workers` option is not used, the number of workers used is the
+provided for the system in the spread.yaml which is 1 by default.
 
 <a name="passwords">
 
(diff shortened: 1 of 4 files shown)


mvo5:
This is interesting - we may instead use a way to specify "limit-to-a-single-one" to avoid side-effects like creating thousands of machines (e.g. --single-worker).

sergiocazzolato:
@mvo @niemeyer I updated the change to allow the new option -single-worker as it was requested on Frankfurt

README.md
```diff
@@ -16,6 +16,7 @@ Spread
 [Fast iterations with reuse](#reuse)  
 [Debugging](#debugging)  
 [Repeating tasks](#repeating)
+[Define number of Workers](#workers)
```
mvo5, line 19: This (and the bits below) does not match what is now implemented. Please update or remove.

spread/project.go
```diff
@@ -854,6 +854,16 @@ func (p *Project) Jobs(options *Options) ([]*Job, error) {
 		return nil, fmt.Errorf("remote project path must be absolute and not /: %s", p.RemotePath)
 	}
 
+	// In case the number of workers set in the options is bigger than 0,
```
mvo5, line 857: This comment is slightly misleading now.

tests/workers/task.yaml
```diff
@@ -0,0 +1,12 @@
...
+prepare: |
+    if [ ! -f .spread-reuse.yaml ]; then
+        touch /run/spread-reuse.yaml
+        ln -s /run/spread-reuse.yaml .spread-reuse.yaml
+    fi
+
+execute: |
+    # Check it is possible to use just 1 worker
+    spread -single-worker lxd:ubuntu-16.04:checks/ &> task1.out
+    [ "$(grep -c Allocating task1.out)" -eq 1 ]
```
mvo5, line 12: Can we have two tests here ? One that checks that the "grep -c" count is one with single worker and one that it is greater than 1 without single-worker? To ensure the option is really having an effect?

README.md
sergiocazzolato: done

spread/project.go
sergiocazzolato: done

tests/workers/task.yaml
sergiocazzolato: Nice catch, but it is not possible yet, I'll implement this test once we can use the google backend for the tests because with lxd we always use 1 worker.  
The pr with the change to start using gitbuh actions (#107) will allow to test that.

sergiocazzolato: Sure, thanks

PR #82, opened 2019-07-27
Introducing the tag for the test which allow to filter tests to execute

sergiocazzolato:
The tests could have a list of tags defined which then can be used to
filter the tests.

For example, if a test has a tad called offline, then when executing
"spread -tag offline <path>" just the tests with that specific tag will
be executed.

### DIFF
--- a/cmd/spread/main.go
+++ b/cmd/spread/main.go
@@ -33,7 +33,9 @@ var (
 	residue        = flag.String("residue", "", "Where to store residual data from tasks")
 	seed           = flag.Int64("seed", 0, "Seed for job order permutation")
 	repeat         = flag.Int("repeat", 0, "Number of times to repeat each task")
+	tag            = flag.String("tag", "", "Filter tests which match with the tag specified")
 	garbageCollect = flag.Bool("gc", false, "Garbage collect backend resources when possible")
+
 )
 
 func main() {
@@ -95,6 +97,7 @@ func run() error {
 		Residue:        *residue,
 		Seed:           *seed,
 		Repeat:         *repeat,
+		Tag:            *tag,
 		GarbageCollect: *garbageCollect,
 	}
 
--- a/spread/project.go
+++ b/spread/project.go
@@ -349,6 +349,7 @@ type Task struct {
 
 	Priority OptionalInt
 	Manual   bool
+	Tags     []string
 }
 
 func (t *Task) String() string { return t.Name }
@@ -984,6 +985,19 @@ func (p *Project) Jobs(options *Options) ([]*Job, error) {
 		}
 		if !manualTasks && job.Task.Manual {
 			continue
+		}
+
+		// Check the tag in the jon in case a tag is specified in the options
+		if options.Tag != "" {
+			tagMatch := false
+			for tag := range job.Task.Tags {
+				if options.Tag == job.Task.Tags[tag] {
+					tagMatch = true
+				}
+			}
+			if !tagMatch {
+				continue
+			}
 		}
 		jobs = append(jobs, job)
 		backends[job.Backend.Name] = true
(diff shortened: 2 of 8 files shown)


sergiocazzolato:
Docs missing. Are gonna be added once the idea is validated.

spread/project.go
```diff
@@ -985,6 +986,19 @@ func (p *Project) Jobs(options *Options) ([]*Job, error) {
 		if !manualTasks && job.Task.Manual {
 			continue
 		}
+
+		// Check the tag in the jon in case a tag is specified in the options
+		if options.Tag != "" {
+			tagMatch := false
+			for tag := range job.Task.Tags {
```
chipaca: when you loop like this you're not getting the tag but the index, so you'd normally have that written as
```go
for i := range job.Task.Tags {
```
or
```go
for _, tag := range job.Task.Tags {
```
In this case I think you want the latter, ...

spread/project.go
```diff
@@ -985,6 +986,19 @@ func (p *Project) Jobs(options *Options) ([]*Job, error) {
...
+				if options.Tag == job.Task.Tags[tag] {
```
chipaca: ...as this then becomes
```go
if options.Tag == tag {
```

spread/project.go
```diff
@@ -985,6 +986,19 @@ func (p *Project) Jobs(options *Options) ([]*Job, error) {
...
+					tagMatch = true
```
chipaca: here you might as well `break`

cmd/spread/main.go
```diff
@@ -33,7 +33,9 @@ var (
 	residue        = flag.String("residue", "", "Where to store residual data from tasks")
 	seed           = flag.Int64("seed", 0, "Seed for job order permutation")
 	repeat         = flag.Int("repeat", 0, "Number of times to repeat each task")
+	tag            = flag.String("tag", "", "Filter tests which match with the tag specified")
 	garbageCollect = flag.Bool("gc", false, "Garbage collect backend resources when possible")
+
```
chipaca: spurious space

spread/project.go
```diff
@@ -985,6 +986,19 @@ func (p *Project) Jobs(options *Options) ([]*Job, error) {
...
+				}
+			}
+			if !tagMatch {
```
chipaca: but, actually, note there's a `contains()` defined in `lxd.go` so drop the whole loop and if and instead simply do
```go
if options.Tag != "" && !contains(job.Task.Tags, options.Tag) {
    continue
}
```

cmd/spread/main.go
sergiocazzolato: removed

spread/project.go
sergiocazzolato: done

spread/project.go
sergiocazzolato: done

spread/project.go
sergiocazzolato: done

spread/project.go
sergiocazzolato: @chipaca in this case it is not importing the lxd.go file, should I create a utils.go and move that function there instead?

spread/project.go
sergiocazzolato: @chipaca thanks for the review

spread/project.go
chipaca: @sergiocazzolato it's in the same package so it's already "there". I don't think moving it to utils.go would be useful (it's very small).

sergiocazzolato: Ah, nice, thanks for the answer!

sergiocazzolato:
@chipaca Documentation included now

mvo5 (approved):
Thanks, I think this is fine as a start. It seems like tags are something very useful and I can imagine a richer syntax in the future, maybe `^` to exclude tags and allow comma separation. But this PR seems to be fine as the first iteration on this journey.

PR #85, opened 2019-08-13
spread/client: workaround bash 4.3 subshell errexit issues

bboozzoo:
Parts of the stage script (prepare, restore..) are assembled into one script
using subshells. REBOOT and ERROR helpers make the shell exit with special
status code 213 which also counts as an error exit status. Normally, we expect
the error exit status to bubble up and be observed by the SSH session runner.

However, due to bug in bash 4.3, this does not happen on systems using that
particular version. The set -e switch causes upper subshell to exit with status
1, instead of forwarding the exit status of the inner shell, provided the inner
one is not the last.

Thus an example script like this:

```
  set -eu
  set -x
  ( # stage

  (
    # snippet, eg. restore
    REBOOT
  )

  (
   # snippet, eg. restore-each
  )
  ) < /dev/null
```

Exits with status 1, instead of 213. This can be reproduced with minimal setup:

```
restore-each: |
    echo foo

restore: |
    echo doing reboot
    REBOOT
```

On a system with bash 4.3 (eg. Ubuntu 16.04), REBOOT is not executed during
restore.

ping @mvo5 @zyga @chipaca

### DIFF
--- a/spread/project.go
+++ b/spread/project.go
@@ -440,7 +440,7 @@ func join(scripts ...string) string {
 		}
 		buf.WriteString("(\n")
 		buf.WriteString(script)
-		buf.WriteString("\n)")
+		buf.WriteString("\n) || exit $?;\n")
 	}
 	return buf.String()
 }
--- a/spread/client.go
+++ b/spread/client.go
@@ -467,8 +467,31 @@ func (c *Client) runPart(script string, dir string, env *Environment, mode outpu
 	if e, ok := err.(*ssh.ExitError); ok {
 		debugf("Script exit status: %v", e.ExitStatus())
 		if e.ExitStatus() == 213 {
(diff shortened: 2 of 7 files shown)


anonymouse64:
What's the status of this?

anonymouse64:
Any news here?

spread/project.go
```diff
@@ -440,7 +440,7 @@ func join(scripts ...string) string {
 		}
 		buf.WriteString("(\n")
 		buf.WriteString(script)
-		buf.WriteString("\n)")
+		buf.WriteString("\n) || exit $?;\n")
```
mvo5, line 443: Having a comment and ideally a bug reference or a commit would be good, i.e. it's not obvious without that why this is needed or wanted.

PR #88, opened 2019-10-08
introduce NOMATCH assertion helper

cjp256:
Using MATCH -v works fine for single-line input, but fails
to make an equivalent assertion on multi-line input.  Add
a new helper which will error if match is found.

Signed-off-by: Chris Patterson <chris.patterson@canonical.com>


cjp256:
Test failure looks unrelated  https://travis-ci.org/snapcore/spread/builds/595297526#L397

Maybe just needs a restart?

PR #89, opened 2019-10-10
runner: pass --exclude-vcs-ignores to tar

zyga:
This allows to use .gitignore and other similar files if they are
present in the current working directory. Note that this is not a
breaking change in any way. The following paragraphs explain why.

When tar is invoked with --exclude-vcs-ignores and the "." as the list
of things to pack, then the exclude logic works correctly. When tar is
invoked with the same option but with each directory and file spelled
out explicitly then --exclude-vcs-ignores does nothing.

This can be tested as follows. Create a .gitignore with *.o rule. Create
a directory subdir with foo.o inside and run:

    tar -c --exclude-vcs-ignores subdir | tar -t | grep -E '\.o$'
    tar -c --exclude-vcs-ignores .gitignore subdir | tar -t | grep -E '\.o$'
    tar -c --exclude-vcs-ignores . | tar -t | grep -E '\.o$'

The first and second commands both ignore the rules in the .gitignore
file. The last command works as people might expect.

Spread has a notion of include list, which contains the set of project
files to send for testing. Spread computes the include list if one is
not defined in spread.yaml. The computed value contains all the files
and directories in the root of the project with the files matching
".spread-reuse.*" filtered-out. Curiously the very tar command this
patch is changing also explicitly excludes the same pattern, which seems
redundant.

The consequence is that for the new options to actually do something in
an existing spread project one needs to explicitly modify the
spread.yaml to read:

    include: [.]

This modification to a project can be coupled with the removal of the
typical exclude list, since it can be typically moved entirely to
.gitignore or similar.

Signed-off-by: Zygmunt Krynicki <zygmunt.krynicki@canonical.com>

### DIFF
--- a/spread/runner.go
+++ b/spread/runner.go
@@ -277,7 +277,11 @@ func (r *Runner) prepareContent() (err error) {
 		return fmt.Errorf("cannot remove temporary content file: %v", err)
 	}
 
-	args := []string{"c", "--exclude=.spread-reuse.*"}
+	args := []string{
+		"-c",
+		"--exclude=.spread-reuse.*",
+		"--exclude-vcs-ignores",
+	}
 	if r.project.Repack == "" {
 		args[0] = "cz"
 	}


spread/runner.go
```diff
@@ -277,7 +277,12 @@ func (r *Runner) prepareContent() (err error) {
 		return fmt.Errorf("cannot remove temporary content file: %v", err)
 	}
 
-	args := []string{"c", "--exclude=.spread-reuse.*"}
+	args := []string{
+		"-c",
+		"--exclude=.spread-reuse.*",
+		"--exclude-vcs-ignores",
+		"--exclude-vcs",
```
bboozzoo: This one may cause problems when one tries to use the information from the actual vcs in a task. A use case could be building a package locally in a prepare and consuming the output of `git describe`.

zyga: I can drop this. It's not essential for `--exclude-vcs-ignores` from working

zyga:
```suggestion
```

zyga:
Force pushed to fix typos in the original message, sorry, didn't notice my spellchecker was off.

zyga:
The corresponding change to snapd's spread.yaml is https://github.com/snapcore/snapd/pull/7585

mvo5:
So if I read this correctly you are saying that adding:
```
include: [.]
```
means "please exclude based on my .gitignore" (or .bzrignore or whatever is used) - is that correct?
If I understand this correctly I think we want a better way to express this, this seems to be not very intuitive.

zyga:
@mvo5 yes, that's correct.

We can certainly include another project option or even make `include: [.]` the default, where it would just kick-in for every existing project. I'm happy to explore those options.

anonymouse64:
From what @mvo5 said, I agree this behavior is a bit confusing, what about something like:

```yaml
project-exclude-method: vcs
```

which then triggers the inclusion of `--exclude-vcs-ignores` to the tar command, and makes it illegal to include other `exclude` and `include` options in the yaml. 

That would also allow the eventual implementation or design of

```yaml
external-exclude-method: git
```

for specific git behavior, etc. I fear that this implementation however would require refactoring in spread to allow for different methods (in the Go code) of how the archive is created, with the default and the `vcs` method being two different implementations.

zyga:
@anonymouse64 that's all doable, I think personally I would go for good defaults where it just works out of the box but you can set exclude: null to explicitly opt out. I think it would be the most useful arrangement.

anonymouse64:
The more I think about it, the more it feels like that by us adding `--exclude-vcs-ignores` to `tar` by default is adding weird behavior around that. I don't think it's obvious that by default when using spread you are ignoring vcs files, unless you specify something else (via `include` or `exclude`) in the yaml. How would you specify that you want to ignore vcs files and also ignore some other file for spread with this?

Furthermore, this whole weirdness around tar when you have other excludes or includes in the arguments to tar make including `--exclude-vcs-ignores` not do anything feels like a bug, or at the very least unexpected behavior, so rather than have us inherit that weird behavior I think it's cleaner for us to make you opt-in to the weirdness.

bboozzoo:
Opt-in sounds ok to me, but I'm not sure about adding a new, yet still opaque, configuration settings. I'd rather prefer to have full control over tar arguments (and compression method, but that's another story). Why not have a top level list of exclude options that can be passed to tar, eg:
```
exclude-args:
  - --exclude-caches
  - --exclude-vcs-ignores
```
Obviously only options that start with `--exclude-` prefix would be allowed.

@zyga have you tried achieving the same result by repacking the sources? I wonder if this would work for snapd:
```
repack: |
  tmpdir=$(mktemp -d)
  trap "rm -rf "$tmpdir"" EXIT
  tar -x -C "$tmpdir" <&3
  tar -c -C "$tmpdir" --exclude-vcs-ignores . >&4
```

zyga:
@bboozzoo I didn't, the whole reason for this PR is to have a discussion to improve spread. We can work around this snapd side with an elaborate exclude list or perhaps with your idea. My intent was to improve the tool itself, for everyone. Good tools have good defaults.

anonymouse64:
> I'd rather prefer to have full control over tar arguments

Putting on my fake architect hat, I'd say that this is _too specific_ to the tar implementation of how the archive is created. For example, maybe what someone wants is to use squashfs or zip or even 7zip, etc. to create an archive. 

Then again, maybe this is fine because spread only works with tar format, in which case a `exclude-args` makes sense but maybe as `pack-tar-args` to make it clear what the args are for.

bboozzoo:
> Putting on my fake architect hat, I'd say that this is too specific to the tar implementation of how the archive is created. For example, maybe what someone wants is to use squashfs or zip or even 7zip, etc. to create an archive.

IDK, tar is assumed all the way through the stack. If you want squashfs or other format, you can only do that via `repack`, which still has to produce a tar at the output. As I suggested in the previous comment, one can also use repack section to apply `--exclude-vcs-ignores`, though you'd have to use an intermediate directory to extract the package and pack it up again.

I think I'm leaning towards closing the PR after all. I belive the problem we had got addressed in https://github.com/snapcore/snapd/pull/7587

zyga:
I'm happy to explore how to keep this as an option. I think for usability having to maintain two exclusion lists for software development projects is a very annoying property. I understand that it may not be desired in some specific cases as well that the current "enabling" mechanism is a bit silly (it's a side effect really).

I'd like to consider changing it so that `include: [.]` is no longer required and that git exclusions are automatic default with an opt-out.

anonymouse64:
> I'd like to consider changing it so that include: [.] is no longer required and that git exclusions are automatic default with an opt-out.

This seems like the smartest approach, I'd rather have an explicit design choice changing this for everybody by default than introduce a "sometimes on" feature that relies on this confusing tar option behavior.

zyga:
I've opened https://github.com/snapcore/spread/pull/90 that will allow us to introduce the new option for filtering VCS ignores. Please review that one.

zyga:
I'll rework this change to make the option configurable. I'll open a new PR for that and refer to this one for extra history.

PR #90, opened 2019-10-21
runner: remove filterDir, default include to "."

zyga:
The include list was defaulting to all the files and directories in the
project, except for files matching .spread-reuse-*. The list of files is
then forwarded to tar, along with the same exclude rule.

This significantly simplifies the invocation of tar and opens up the
possibility of using .gitignore, which requires tar to read the root of
the directory to work.

Two new spread tests check this behavior.

### DIFF
--- a/spread/runner.go
+++ b/spread/runner.go
@@ -289,10 +289,7 @@ func (r *Runner) prepareContent() (err error) {
 	}
 	include := r.project.Include
 	if len(include) == 0 {
-		include, err = filterDir(r.project.Path)
-		if err != nil {
-			return fmt.Errorf("cannot list project directory: %v", err)
-		}
+		include = []string{"."}
 	}
 	args = append(args, include...)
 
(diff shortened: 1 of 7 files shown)


zyga:
@anonymouse64 no, it doesn't since the tar code already removes them so they are never sent, regardless of this patch. I only de-duplicated the code that handles that. Note that the spread test shows that .spread-reuse.test is never sent, regardless of what spread.yaml says.

spread/runner.go
```diff
@@ -289,10 +289,7 @@ func (r *Runner) prepareContent() (err error) {
 	}
 	include := r.project.Include
 	if len(include) == 0 {
-		include, err = filterDir(r.project.Path)
```
niemeyer: I'm missing the actual motivation of the change. It looks like these are not doing the same thing. Description says it simplifies.. but it doesn't talk about the rationale and consequences for the actual behavior change.

niemeyer:
Per discussion today, this is not just simplifying it. This is changing behavior, and the output tar file actually holds different paths with a prefix of "./". Besides being somewhat ugly, this may also affect the patterns applied via inclusion and exclusion rules.

I'm tentatively closing this on the premise that it doesn't actually fix or improve anything else either. Please feel free to bring it up for discussion if you feel that's not correct.

PR #91, opened 2019-10-21
tests: Fix adhoc test

sergiocazzolato:
Always take first backend which is lxd because depending on the previous
tests could be more than one backend use in teh .spread-reuse.yaml file

The test fails when the qemu test is executed before

google:ubuntu-18.04-64 .../tests/adhoc# cat .spread-reuse.yaml
backends:
  lxd:
    systems:
    - ubuntu-16.04:
        password: 9facef8d38c59bba
        address: 10.4.66.64
        data:
          name: spread-1-ubuntu-16-04
  qemu:
    systems:
    - ubuntu-16.04:
        username: ubuntu
        password: ubuntu
        address: localhost:59327
        data:
          pid: 7046

### DIFF
--- a/tests/adhoc/task.yaml
+++ b/tests/adhoc/task.yaml
@@ -11,8 +11,8 @@ execute: |
     spread -reuse lxd:
 
     export ADHOC_USERNAME=root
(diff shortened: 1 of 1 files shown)


niemeyer:
This is now fixed by a PR from @zyga.

PR #93, opened 2019-10-23
snap: update to classic confinement

cjp256:


mvo5:
Thanks for the PR - could you please elaborate a bit why the "gcc" build-package was added?

cjp256:
@mvo5 When building, the following error occurs:
```
go build github.com/snapcore/spread/cmd/spread
go build -ldflags -linkmode=external github.com/snapcore/spread/cmd/spread
# github.com/snapcore/spread/cmd/spread
/snap/go/5364/pkg/tool/linux_amd64/link: running gcc failed: exec: "gcc": executable file not found in $PATH
```

This appears to be due to a workaround:
https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/go.py#L269

I would say that there is a strong case that the snapcraft go plugin should add gcc to build-packages for classic snaps as long as that workaround is in play. @sergiusens what do you think?

While here, I'd also ask if `classic` mode is desirable?  Is there a reason anyone thinks that this may not be capable of running with strict with a proper set of plugs?  My original thought was that most developer tools seem to use classic, and it was either that or devmode.   But I'd be willing to try to get it running as strict if you think it's doable/appropriate without significant changes to spread @mvo5 .

sergiusens:
You either need classic or either bundle the lxc client or implement the API in spread.

PR #94, opened 2019-12-11
qemu: invoke qemu in a portable way

zyga:
The kvm binary is non-standard packaging tweak in available in the
Debian world. To work on Fedora and all the other systems as well, we
can just invoke qemu-system-x86_64 with the -enable-kvm switch.


zyga:
For context, `kvm` is a Debian script that contains literally this:

```
#!/bin/sh
exec qemu-system-x86_64 -enable-kvm "$@"
```

PR #95, opened 2019-12-19
spread: add support to define a custom bios with the qemu backend

mvo5:
For some tests it is important to customize the bios that qemu
uses. One example is the Ubuntu Core 20 project. For the spread
testing there we need a qemu that uses the OVMF uefi bios.

The location of this bios is different on all distributions so
defining it in the spread.yaml seems not appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -827,6 +827,10 @@ see the `-pass` command line option).
 The QEMU backend is run with the `-nographic` option by default. This
 may be changed with `export SPREAD_QEMU_GUI=1`.
 
+The bios that qemu uses can be changed via the SPREAD_QEMU_BIOS
+environment variable. To switch to an UEFI bios on Ubuntu
+`export SPREAD_QEMU_BIOS=/usr/share/OVMF/OVMF_CODE.ms.fd` can be used.
+
 Note that at the moment QEMU is run via the `kvm` script, which enables
 the KVM performance optimizations for the local architecture. This will
 not work for other architectures, though. This problem may be easily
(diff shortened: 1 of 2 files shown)


spread/qemu.go
```diff
@@ -118,6 +118,10 @@ func (p *qemuProvider) Allocate(ctx context.Context, system *System) (Server, er
 	if os.Getenv("SPREAD_QEMU_GUI") != "1" {
 		cmd.Args = append([]string{cmd.Args[0], "-nographic"}, cmd.Args[1:]...)
 	}
+	if biosPath := os.Getenv("SPREAD_QEMU_BIOS"); biosPath != "" {
+		cmd.Args = append([]string{cmd.Args[0], "-bios", biosPath}, cmd.Args[1:]...)
+	}
```
cmatsuoka: We may need to use `-pflash` later when working with secure boot to be able to enroll our keys and certificates to our nvram file (which we'll also need to specify).

README.md
```diff
@@ -827,6 +827,10 @@ see the `-pass` command line option).
 The QEMU backend is run with the `-nographic` option by default. This
 may be changed with `export SPREAD_QEMU_GUI=1`.
 
+The bios that qemu uses can be changed via the SPREAD_QEMU_BIOS
+environment variable. To switch to an UEFI bios on Ubuntu
+`export SPREAD_QEMU_BIOS=/usr/share/OVMF/OVMF_CODE.ms.fd` can be used.
```
mvo5: This can't be used as is because the bios location is a property of the host. But the images used can be both efi/non-efi and some efi images may need to be booted with a different efi bios. So one global options like this is not granular enough.

alfonsosanchezbeato: Have you considered adding a `bios:` string when necessary in `backends.systems` instead?

alfonsosanchezbeato: Have you considered adding an optional `bios:` string for `backends.systems` in `spread.yaml` instead?

mvo5:
Closing this in favor of https://github.com/snapcore/spread/pull/116

PR #96, opened 2019-12-19
spread: add support for system specific "flags" and use in qemu

mvo5:
This commit adds a concept of system specific flags. This is useful
to enable certain properties on the given system. The first use-case
is to switch virtio on for a specific system. This will be useful
for UC20 where the UEFI grub bootloader is very very slow (minutes)
when loading the kernel in non-virtio mode.

The flags concept will probably be useful for other use-cases as
well. One example is to add "uefi" to google backend so that
systems that have this flag are created with UEFI enabled.

An alternative approach would be to just check an environment variable
like SPREAD_QEMU_VIRTIO=1. Enabling this globally is risky because
the device name change from /dev/sda to /dev/vda so tests that look at
those will need updating.

### DIFF
--- a/README.md
+++ b/README.md
@@ -813,6 +813,7 @@ backends:
             - ubuntu-16.04:
                 username: ubuntu
                 password: ubuntu
+                flags: [virtio]
 ```
 
 For this example to work, a QEMU image must be made available under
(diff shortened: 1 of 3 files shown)


mvo5:
I think we really don't want this, it's the wrong approach.

PR #97, opened 2020-01-08
spread: deal with crashing qemu in waitPortUp()

mvo5:
Someones qemu crashes or dies while starting up. If that happens
spread does not detected it and keeps trying to connect to the
port until it times out. This leaves the user in the dark why
the connection did not happen.

This commit improves this in two ways:
1. Show the stdout/stderr output of qemu on the terminal to
   ensure that users are aware of any issues there.
2. Add a check in waitPortUp() to detect if the command that
   is responsible for the setting up of the port exited in an
   unexpected way.

### DIFF
--- a/spread/client.go
+++ b/spread/client.go
@@ -1011,7 +1011,7 @@ func outputErr(output []byte, err error) error {
 	return err
 }
 
-func waitPortUp(ctx context.Context, what fmt.Stringer, address string) error {
+func waitPortUp(ctx context.Context, what fmt.Stringer, address string, cmd *exec.Cmd) error {
 	if !strings.Contains(address, ":") {
 		address += ":22"
 	}
@@ -1029,6 +1029,14 @@ func waitPortUp(ctx context.Context, what fmt.Stringer, address string) error {
 			conn.Close()
 			break
 		}
+		if cmd != nil {
+			var wstatus syscall.WaitStatus
+			wpid, err := syscall.Wait4(cmd.Process.Pid, &wstatus, syscall.WNOHANG, nil)
+			if err != nil || wpid != 0 {
+				return fmt.Errorf("process exited unexpectedly while waiting for address %s (wstatus=%v)", address, wstatus)
+			}
+		}
+
 		select {
 		case <-retry.C:
 		case <-relog.C:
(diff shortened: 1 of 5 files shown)


spread/client.go
```diff
@@ -1011,7 +1011,7 @@ func outputErr(output []byte, err error) error {
 	return err
 }
 
-func waitPortUp(ctx context.Context, what fmt.Stringer, address string) error {
+func waitPortUp(ctx context.Context, what fmt.Stringer, address string, cmd *exec.Cmd) error {
```
mvo5: Instead of adding a extra parameter here we need to teach the call site to interrupt ctx when cmd dies. So a go-routine to monitor cmd and cancel the context.

spread/qemu.go
```diff
@@ -125,6 +132,29 @@ func (p *qemuProvider) Allocate(ctx context.Context, system *System) (Server, er
...
+	// that is used by waitPortUp
+	ctx, cancelFn := context.WithCancel(ctx)
+	portUpDoneCh := make(chan struct{})
+	go func() {
+		var retry = time.NewTicker(500 * time.Millisecond)
+		for {
+			var wstatus syscall.WaitStatus
+			wpid, err := syscall.Wait4(cmd.Process.Pid, &wstatus, syscall.WNOHANG, nil)
+			if err != nil || wpid != 0 {
+				print("qemu exited unexpectedly: %v", wstatus)
```
mvo5, line 145: This needs to be "printf"

spread/qemu.go
```diff
@@ -125,6 +132,29 @@ func (p *qemuProvider) Allocate(ctx context.Context, system *System) (Server, er
...
+	// watch if qemu comes up as expected and if not cancel the context
+	// that is used by waitPortUp
+	ctx, cancelFn := context.WithCancel(ctx)
+	portUpDoneCh := make(chan struct{})
+	go func() {
+		var retry = time.NewTicker(500 * time.Millisecond)
+		for {
+			var wstatus syscall.WaitStatus
+			wpid, err := syscall.Wait4(cmd.Process.Pid, &wstatus, syscall.WNOHANG, nil)
+			if err != nil || wpid != 0 {
```
zyga, line 144: I think you want `wpid == -1` instead

spread/qemu.go
```diff
@@ -125,6 +132,29 @@ func (p *qemuProvider) Allocate(ctx context.Context, system *System) (Server, er
...
 
+	// watch if qemu comes up as expected and if not cancel the context
+	// that is used by waitPortUp
+	ctx, cancelFn := context.WithCancel(ctx)
+	portUpDoneCh := make(chan struct{})
+	go func() {
+		var retry = time.NewTicker(500 * time.Millisecond)
+		for {
+			var wstatus syscall.WaitStatus
+			wpid, err := syscall.Wait4(cmd.Process.Pid, &wstatus, syscall.WNOHANG, nil)
```
mvo5, line 143: Use go blocking Wait - but if anything goes wrong from portUp kill the process.

PR #98, opened 2020-02-11
Adding maxResults parameters when requesting images from google backend

sergiocazzolato:
The ideas of this change is to get up to 500 images from google backend.
This is a workaound for the paginatino issue.

The problem with the order parameter is that because an issue in the
gcloud api, when the orderBy parameter is used, the maxResults does not
work.

As the results are not so many, the idea is to manually order the
results.

I already submitted an issue explaining that bad behaviour:
https://issuetracker.google.com/issues/149183460

### DIFF
--- a/spread/google.go
+++ b/spread/google.go
@@ -256,14 +256,11 @@ Outer:
 func (p *googleProvider) getLastCreatedImage(images []googleImage) (image googleImage) {
 	var lastImageCreated googleImage
 	lastImageCreated = images[0]
-	printf("Initial time %v", lastImageCreated)
 	for i := range images[1:] {
-		printf("Parsing time %v", images[i])
 		if lastImageCreated.CreationTimestamp.Before(images[i].CreationTimestamp) {
 			lastImageCreated = images[i]
 		}
 	}
-	printf("Final time %v", lastImageCreated)
 	return lastImageCreated
 }
 
--- a/spread/google.go
+++ b/spread/google.go
@@ -253,7 +253,7 @@ Outer:
 	return true
 }
 
-func (p *googleProvider) getLastCreatedImage(images []googleImage) (image googleImage) {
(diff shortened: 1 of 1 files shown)


spread/google.go
```diff
@@ -252,8 +253,20 @@ Outer:
 	return true
 }
 
+func (p *googleProvider) getLastCreatedImage(images []googleImage) (image googleImage) {
```
anonymouse64:
```suggestion
func (p *googleProvider) newestCreatedImage(images []googleImage) (image googleImage) {
```

spread/google.go
```diff
@@ -306,18 +319,26 @@ func (p *googleProvider) image(system *System) (image *googleImage, family bool,
 			image := &images[i]
 			if image.Family == lookup.Family {
 				debugf("Family match: %#v matches %#v", lookup, image)
-				return image, true, nil
+				matchingImages = append(matchingImages, *image)
 			}
 		}
+		if len(matchingImages) > 0 {
+			image := p.getLastCreatedImage(matchingImages)
```
anonymouse64:
```suggestion
			image := p.newestCreatedImage(matchingImages)
```

spread/google.go
```diff
@@ -352,29 +373,37 @@ func (p *googleProvider) projectImages(project string) ([]googleImage, error) {
...
+			Project: 			project,
+			Name:    			item.Name,
+			Family:  			item.Family,
+			Terms:   			toTerms(item.Description),
+			CreationTimestamp: 	parsedCreationTimestamp,
 		})
 	}
 
+
+
```
anonymouse64: this whitespace can be deleted

sergiocazzolato: done

spread/google.go
sergiocazzolato: done

spread/google.go
sergiocazzolato: done

PR #100, opened 2020-03-03
spread.yaml: remove `image: ubuntu-1804-64-virt-enabled` line

mvo5:
This should make travis happy.


mvo5:
This looks like it was not the right fix.

PR #104, opened 2020-04-16
Adding option for systems created on google to start with secure boot enabled

sergiocazzolato:
This is a feature to enable the secure boot option in google instances.

It is needed for running some uc20 tests because the key sealing in the
encryption tests needs secure boot to be enabled

The configuration in the spread.yaml should be:
systems:
    - ubuntu-20.04-64:
         workers: 6
         secureboot: true

### DIFF
--- a/spread/google.go
+++ b/spread/google.go
@@ -433,6 +433,15 @@ func (p *googleProvider) createMachine(ctx context.Context, system *System) (*go
 		diskParams["diskSizeGb"] = int(system.Storage / gb)
 	}
 
+	secureBootParams := googleParams{}
+	if system.SecureBoot {
+		secureBootParams = googleParams{
+			"enableSecureBoot": true,
+			"enableVtpm": true,
+			"enableIntegrityMonitoring": true,
+		}
+	}
+
 	params := googleParams{
 		"name":        name,
 		"machineType": "zones/" + p.gzone() + "/machineTypes/" + plan,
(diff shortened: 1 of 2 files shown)


spread/project.go
```diff
@@ -122,6 +122,9 @@ type System struct {
 	// Only for Linode and Google so far.
 	Storage Size
 
+	// Only for Google so far.
+	SecureBoot	bool
```
niemeyer, line 126: Every other option we have with multiple words use dashes to separate them. We should do this here as well.

spread/google.go
```diff
@@ -433,6 +433,15 @@ func (p *googleProvider) createMachine(ctx context.Context, system *System) (*go
 		diskParams["diskSizeGb"] = int(system.Storage / gb)
 	}
 
+	secureBootParams := googleParams{}
```
niemeyer, line 436: This allocates a map ...

spread/google.go
```diff
@@ -433,6 +433,15 @@ func (p *googleProvider) createMachine(ctx context.Context, system *System) (*go
...
+	if system.SecureBoot {
+		secureBootParams = googleParams{
```
niemeyer, line 438: ... and this immediately reallocates it. It's not a big deal, but avoiding it is also trivial.

cmatsuoka:
PR #105 tries to address the issues raised in this review while fixing a problem related to empty `shieldedInstanceConfiguration` in certain Fedora and Amazon Linux images.

PR #105, opened 2020-05-20
google: only configure shielded instance when secure boot is enabled

cmatsuoka:
Set shielded instance configurations only when secure boot is enabled,
using an empty configuration is not well-received in some systems. Also
change the option to enable secure boot to "secure-boot" to be consistent
with other options that use dash between words.


niemeyer:
Sorry for the confusion. I've done that change back in 0e3ac770255ab1940656c917f1b585cc2f1d4a75 but apparently didn't push it.

PR #107, opened 2020-05-28
Use GitHub actions to test spread

sergiocazzolato:
This is the first step to use github actions, then the idea is to start testing the google backend and other backend as well.

This change also makes a re-org spread.yaml and the test suites.

Once github actions is enabled in the project the tests can be automatically executed. 
Currently tests with github actions are being executed in: https://github.com/sergiocazzolato/spread/pull/1


spread.yaml
```diff
@@ -11,11 +11,43 @@ environment:
...
     GOPATH: $GOHOME
+    PROJECT_PATH: $GOHOME/src/github.com/snapcore/spread
     PATH: $GOHOME/bin:$PATH
     CACHE_DISABLED: $(HOST:[ "$SEND_CACHE" = 1 ] && echo false || echo tests/cache)
 
 backends:
     google:
+        key: "$(HOST: echo $SPREAD_GOOGLE_AUTH_JSON)"
+        location: computeengine/us-east1-b
+        halt-timeout: 2h
```
zyga, line 22: Do we need two hours? Perhaps less is okay?

spread.yaml
```diff
@@ -11,11 +11,43 @@ environment:
...
+        systems:
+            - ubuntu-14.04-64:
```
zyga, line 24: I would probably trim this list to include: ubuntu-latest, fedora-latest, opensuse-latest and maybe debian. The rest seem like waste as we're not actually using them for development.

spread.yaml
```diff
@@ -38,48 +70,57 @@ exclude:
 path: /home/test/src/github.com/snapcore/spread
 
 suites:
-    tests/: 
+    tests/main/:
         summary: Integration tests
+        backends: [google-nested, qemu]
+        prepare: |
+            DEBS="git qemu-kvm lxd tree"
```
zyga, line 77: This is duplicated with what's installed in github actions. You can probably just "apt-get update; apt-get install $DEBS" instead of using the loop below.

spread.yaml
```diff
@@ -38,48 +70,57 @@ exclude:
...
-
-    DEBS="git qemu-kvm lxd tree"
-    if ! dpkg -l $DEBS; then
-        apt update
-        apt install -y $DEBS
-    fi
-
-    # Cache is only available if SEND_CACHE=1 is set.
-    mkdir -p tests/cache
+            # Either use pregenerated spread if exists or build it
```
zyga, line 85: When is spread pre-generated?

spread.yaml
```diff
@@ -38,48 +70,57 @@ exclude:
...
+            if [ -x "$PROJECT_PATH/bin/spread" ]; then
+                mkdir -p "$GOHOME/bin"
+                cp -f "$PROJECT_PATH/bin/spread" "$GOHOME/bin/spread"
+            else
+                # Cache Go deb with: cd tests/cache && godeb download $GOVERSION
```
zyga, line 90: What is the point of this cache?

sergiocazzolato:
Currently tests fail because of the missing google key
https://github.com/sergiocazzolato/spread/runs/2458013693?check_suite_focus=true

sergiocazzolato:
Tests are passing with latest changes in my project:
https://github.com/sergiocazzolato/spread/runs/2795159594?check_suite_focus=true

.github/workflows/test.yaml
```diff
@@ -0,0 +1,49 @@
...
+          go get ./cmd/spread
+          
+      - name: run tests
+        env:
+          SPREAD_GOOGLE_KEY: ${{ secrets.SPREAD_GOOGLE_KEY }}
+        run: |
+          cd ${{ github.workspace }}/src/github.com/snapcore/spread
+          mkdir -p bin
+          cp -f "$GOHOME/bin/spread" ./bin/spread
+          "$GOHOME/bin/spread" google:ubuntu-20.04-64:tests/
```
mvo5: Maybe less specific here? Just ` "$GOHOME/bin/spread" google:` ? So that if we add compat with 22.04 we don't need to change this part?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt install -y xdelta3
+
+      - name: Setup go version
+        uses: actions/setup-go@v2
+        with:
+          # Using the lastest available go version for bionic
```
mvo5:
```suggestion
          # Using the lastest available go version for focal
```

mvo5 (approved):
Looks good! I did two small tweak that I also describe in the review, please let me know if you think those are inappropriate.

PR #113, opened 2021-05-28
Select min cpu platform for google systems

sergiocazzolato:
The idea of this change is to be able to select which is the minimum cpu platform "min-cpu-platform" for a specific system in google backend.

This is needed to avoid some issues that exist because of incompatibilities between kvm and old Xeon servers. 

Available options are:

Intel Xeon E5 (Sandy Bridge): "Intel Sandy Bridge" -> 2011
Intel Xeon E5 v2 (Ivy Bridge): "Intel Ivy Bridge" -> 2012
Intel Xeon E5 v3 (Haswell): "Intel Haswell" -> 2013
Intel Xeon E5 v4 (Broadwell): "Intel Broadwell" -> 2014
Intel Xeon (Skylake): "Intel Skylake" -> 2015
Intel Xeon (Cascade Lake): "Intel Cascade Lake" -> 2019


mvo5 (approved):
Thanks, one small suggestion inline.

spread/project.go
```diff
@@ -123,7 +123,8 @@ type System struct {
 	Storage Size
 
 	// Only for Google so far.
-	SecureBoot bool `yaml:"secure-boot"`
+	SecureBoot     bool   `yaml:"secure-boot"`
+	MinCpuPlatform string `yaml:"min-cpu-platform"`
```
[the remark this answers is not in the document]
mvo5: Added this now.

spread/project.go
```diff
@@ -124,6 +124,9 @@ type System struct {
 
 	// Only for Google so far.
 	SecureBoot bool `yaml:"secure-boot"`
+	// Select the minimum CPU platform on Google GCE, e.g. "Intel Skylake"
+	// See https://cloud.google.com/compute/docs/instances/specify-min-cpu-platform#availablezones
+	MinCpuPlatform string `yaml:"min-cpu-platform"`
```
mvo5, line 129: Feedback from Gustavo: 
- let's use  "CpuFamily" as the name so that it's generic and could also be used for e.g. linode
- map that then in the google backend to the `min-cpu-platform` they are using

mvo5: This is addressed now.

spread/project.go
```diff
@@ -125,6 +125,11 @@ type System struct {
 	// Only for Google so far.
 	SecureBoot bool `yaml:"secure-boot"`
 
+	// Request a specific CPU family, e.g. "Intel Skylake". This follows
+	// the GCE naming today:
```
mvo5, line 129: This comment needs updating, it's only following the naming for GCE when it's working with GCE but on AWS it would be different so this needs to be more generic.

spread/project.go
```diff
@@ -125,6 +125,11 @@ type System struct {
...
+	// https://cloud.google.com/compute/docs/instances/specify-min-cpu-platform#availablezones
+	CpuFamily string `yaml:"cpu-family"`
```
mvo5, line 131: This needs to be capitalized : CPUFamily.

mvo5 (approved):
Once the comments are addressed

PR #114, opened 2021-06-02
Define the plan for a system on google backend

sergiocazzolato:
The idea is to be able to define the plan at the system level.

In case the system has not a plan defined the instance will use the defined for the backend.

In case the backend has not a plan defined the instance will use the default one.


mvo5 (approved):
Thanks, this looks reasonable.

spread/project.go
```diff
@@ -124,6 +124,7 @@ type System struct {
 
 	// Only for Google so far.
 	SecureBoot bool `yaml:"secure-boot"`
+	Plan       string
```
niemeyer, line 127: It's not true that only Google has this, I believe.

niemeyer: Otherwise, it looks reasonable. A unit test about the new and old behavior would also be welcome.

mvo5: I added a unit test now but it does not fully cover old/new because for that it would need to test createMachine() which involves a bunch of mocking. I can do this but I am a bit worried it might be a bit much for a this (simple) PR.

niemeyer (approved):
Looks great, thanks!

PR #116, opened 2021-06-21
spread: add "bios" system option with {legacy,bios,""} options

mvo5:
To boot UC20 in qemu the system needs to be set to boot using
UEFI. However there is currently no way right now to switch a
backend to UEFI or legacy bios mode.
    
This commit adds the option `bios` to the system and (for now)
allows `legacy`, `uefi`. The empty string or a missing option
means the system default should be used.
    
 This is currently only implemented for qemu.

This is the replacement for https://github.com/snapcore/spread/pull/95


spread/qemu.go
```diff
@@ -114,10 +116,18 @@ func (p *qemuProvider) Allocate(ctx context.Context, system *System) (Server, er
...
-	cmd := exec.Command("qemu-system-x86_64", "-enable-kvm", "-snapshot", "-m", strconv.Itoa(mem), "-net", "nic", "-net", fwd, "-serial", serial, "-monitor", monitor, path)
+	cmd := exec.Command(qemuBinary, "-enable-kvm", "-snapshot", "-m", strconv.Itoa(mem), "-net", "nic", "-net", fwd, "-serial", serial, "-monitor", monitor, path)
 	if os.Getenv("SPREAD_QEMU_GUI") != "1" {
 		cmd.Args = append([]string{cmd.Args[0], "-nographic"}, cmd.Args[1:]...)
 	}
+	switch system.Bios {
+	case "legacy", "":
+		// nothing to do, that is the qemu default
+	case "uefi":
+		cmd.Args = append([]string{cmd.Args[0], "-bios", "/usr/share/OVMF/OVMF_CODE.fd"}, cmd.Args[1:]...)
```
zyga, line 127: This will be a problem across distributions, as the name is not standardized.

My advice is as follows:

1) Readlink /usr/share/spread/qemu-efi-firmware
2) If present, that's the name
3) If the link is not present, use a hard-coded default.

This way any packaged version of spread can simply drop a symbolic link and depend on the right firmware package.

spread/qemu.go
```diff
@@ -114,10 +116,18 @@ func (p *qemuProvider) Allocate(ctx context.Context, system *System) (Server, er
 	serial := fmt.Sprintf("telnet:127.0.0.1:%d,server,nowait", port+100)
 	monitor := fmt.Sprintf("telnet:127.0.0.1:%d,server,nowait", port+200)
 	fwd := fmt.Sprintf("user,hostfwd=tcp:127.0.0.1:%d-:22", port)
-	cmd := exec.Command("qemu-system-x86_64", "-enable-kvm", "-snapshot", "-m", strconv.Itoa(mem), "-net", "nic", "-net", fwd, "-serial", serial, "-monitor", monitor, path)
+	cmd := exec.Command(qemuBinary, "-enable-kvm", "-snapshot", "-m", strconv.Itoa(mem), "-net", "nic", "-net", fwd, "-serial", serial, "-monitor", monitor, path)
```
zyga, line 119: Suggestin: I think this would be easier to mock/test if you broke out this whole part as a helper function, `qemuCmd(p *qemuProvider, s *System)`. Then you can simply call that in tests and avoid all the extra complexity in testing.

spread/qemu.go
```diff
@@ -114,10 +116,18 @@ func (p *qemuProvider) Allocate(ctx context.Context, system *System) (Server, er
...
 	if os.Getenv("SPREAD_QEMU_GUI") != "1" {
 		cmd.Args = append([]string{cmd.Args[0], "-nographic"}, cmd.Args[1:]...)
 	}
+	switch system.Bios {
+	case "legacy", "":
+		// nothing to do, that is the qemu default
```
zyga, line 125: I would leave a TODO to warn about empty value some point.

zyga (commented):
I like the feature but there's a problem in how the firmware is located. I've left a remark on one possible way to resolve that problem. I had a look at the older PR and perhaps an even simpler way to resolve it would be to define an environment variable but I think for the purpose of just working out of the box when packaged, my suggestion is (hopefully) better.

spread/qemu.go
mvo5: The default should work on ubuntu/debian/fedora but it seems it will not on suse so I guess that advise makes sense.

spread/qemu.go
mvo5: I like the idea, if I understand you correctly it requires that I'm able to differentiate between unset and set to the empty string which I don't think I can do currently. It would require something similar to "OptionalInt" in project.go. Not a big deal but I would not want to make this PR bigger and harder to review just for this.

spread/qemu.go
zyga: It can be handled with distro-patch as well so that's not all terrible.

spread/qemu.go
zyga: All you need is a pointer to a string in the struct. Then you can do system.Bios == nil vs *system.Bios == "legacy"

spread/qemu.go
```diff
@@ -97,6 +99,16 @@ func systemPath(system *System) string {
 	return os.ExpandEnv("$HOME/.spread/qemu/" + system.Image + ".img")
 }
 
+func ovmfPath() string {
+	// this will work on ubuntu/debian/fedora
+	ovmfPath := "/usr/share/OVMF/OVMF_CODE.fd"
+	// allow packaging to override default
+	if p, err := os.Readlink("/usr/share/spread/qemu-efi-firmware"); err == nil {
```
zyga, line 106: Tests of this will be affected by the host.

zyga (commented):
Almost LGTM

The new `ovmfPath` needs to be testable. Currently host files will (sometimes) break unit tests.

spread/qemu.go
mvo5: Nice catch, fixed now.

spread/qemu.go
mvo5: Let's do this later/in a followup, I want this initial version to be as small and reviewable as possible.

spread/qemu.go
```diff
@@ -10,6 +10,12 @@ import (
 	"golang.org/x/net/context"
 )
 
+var (
+	ovmfDefaultPath = "/usr/share/OVMF/OVMF_CODE.fd"
```
mvo5, line 14: Having a default for ubuntu/debian sounds good. Let's revert it and look at `.spread/qemu/bios/<bios-name>.img` (e.g. uefi.img) first and if nothing is there we use the default ovmfDefaultPath above.

spread/qemu.go
```diff
@@ -10,6 +10,12 @@ import (
...
+	// can be used by distro packaging to point to the right OVMF
+	optionalOvmfPackageSymlink = "/usr/share/spread/qemu-efi-firmware"
```
mvo5, line 16: This one gets removed.

spread/qemu.go
```diff
@@ -97,30 +103,65 @@ func systemPath(system *System) string {
...
+		"-net", "nic",
+		"-net", fwd,
+		"-serial", serial,
+		"-monitor", monitor,
+		path)
 	if os.Getenv("SPREAD_QEMU_GUI") != "1" {
 		cmd.Args = append([]string{cmd.Args[0], "-nographic"}, cmd.Args[1:]...)
 	}
+	switch system.Bios {
+	case "legacy", "":
```
mvo5, line 136: No explicit legacy - empty or "nil"

spread/qemu.go
```diff
@@ -97,30 +103,65 @@ func systemPath(system *System) string {
...
+	// allow packaging to override default
+	if _, err := os.Readlink(optionalOvmfPackageSymlink); err == nil {
+		return optionalOvmfPackageSymlink
 	}
-
-	mem := 1500
-	if p.backend.Memory > 0 {
-		mem = int(p.backend.Memory / mb)
+	// can be used by e.g. the snap
+	if p := os.Getenv("SPREAD_QEMU_OVMF_PATH"); p != "" {
```
mvo5, line 113: This needs to be renamed to something like (strawman) SPREAD_QEMU_FALLBACK_BIOS_PATH that looks at 
$SPREAD_QEMU_FALLBACK_BIOS_PATH/uefi.img

spread/project.go
```diff
@@ -125,6 +125,9 @@ type System struct {
 	// Only for Google so far.
 	SecureBoot bool `yaml:"secure-boot"`
 
+	// Supported are {"legacy","uefi",""}, only for qemu so far.
```
mvo5, line 128: This comment needs updating: we agreed to have no legacy

mvo5: This is done now

PR #117, opened 2021-07-04
Fix OptionalInt.String use of strconv.FormatInt

zyga:
The second argument to strconv.FormatInt is not integer size, but
numeric base. This is reported by go vet, as base cannot be larger than
36.


mvo5 (approved):
Thank you!

PR #121, opened 2021-07-04
google: simplify toTerms function

zyga:
This simplification was suggested by go vet.


mvo5 (approved):
Thank you and good job `go vet`!

mvo5:
I would feel slightly better with a test here but I starred at this for a couple of minutes and could not find a flaw. Also would feel better having integration tests back of course :)

PR #122, opened 2021-07-04
client: simplify append loop in Client.Send

zyga:
This simplification was suggested by go vet.


mvo5 (approved):
Thank you

PR #124, opened 2021-07-04
google,humbox: rename foo to errFoo

zyga:
This is flagged by `go vet` and is in accordance with go error naming conventions.


mvo5 (approved):
Thank you

PR #126, opened 2021-07-04
runner: reduce for+select to for range in runner.Loop

zyga:

### DIFF
--- a/spread/runner.go
+++ b/spread/runner.go
@@ -234,17 +234,14 @@ func (r *Runner) loop() (err error) {
 		}
 	}
 
-	for {
-		select {
-		case <-r.done:
-			r.alive--
-			if r.alive > 0 {
-				debugf("Worker terminated. %d still alive.", r.alive)
-				continue
-			}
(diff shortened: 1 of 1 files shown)


mvo5 (approved):
Thank you, mostly cosmetic but looks smaller/less nested.

spread/runner.go
```diff
@@ -234,17 +234,14 @@ func (r *Runner) loop() (err error) {
...
-			debugf("Worker terminated.")
-			return nil
+	for range r.done {
+		r.alive--
+		if r.alive > 0 {
+			debugf("Worker terminated. %d still alive.", r.alive)
+			continue
 		}
+		debugf("Worker terminated.")
+		return nil
```
mvo5, line 244: Hm, looks like this is not ready. Current go complains:
```
./runner.go:246:1: missing return at end of function
```
we could of course just move the reutrn down but i would feel better I think if we had a a unit test here, given the function I'm not sure that is easy to do though :/

PR #128, opened 2021-07-23
Fix actions workflow and discard workers

sergiocazzolato:
The idea is to use the published spread to trigger the tests execution.

We don't need to build spread in the github action machine because then
spread is built and tested in the target system.

Also it is included a step to discard workers in case it is needed.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -18,32 +18,21 @@ jobs:
           sudo apt-get update
           sudo apt install -y xdelta3
 
-      - name: Setup go version
-        uses: actions/setup-go@v2
-        with:
-          # Using the lastest available go version for focal
-          go-version: 1.13.8
+      - uses: actions/checkout@v2
 
-      - name: Set environment
-        run: |
-          echo "GOHOME=${{ github.workspace }}" >> "$GITHUB_ENV"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -11,39 +11,18 @@ on:
...
-      - name: Build spread
-        run: |
-          cd ${{ github.workspace }}/src/github.com/snapcore/spread
-          go get ./cmd/spread
-          
-      - name: run tests
-        env:
-          SPREAD_GOOGLE_KEY: ${{ secrets.SPREAD_GOOGLE_KEY }}
+      - name: Discard spread workers
+        if: always()
```
mvo5, line 23: I assume we need this to ensure that even for failed spread runs the cleanup runs?

mvo5 (approved):
Thank you

.github/workflows/test.yaml
sergiocazzolato: @mvo5 thanks for the review
Yes, we use this to make sure the machines are discarded in case spread cannot discard them.

mvo5:
Merging this even without an explicit :+1: from Gustavo because it does not touch any of the spread code, just the testing workflow side.

PR #129, opened 2021-07-23
build: move to go modules

mvo5:
This is hopefully straightforward. The only thing that might be controversial is that I hd to change the import in humbox as it appears that the old path was not recognized/found by `go mod`.


zyga (approved):
Looks good.

I'm very happy to see modules adopted.

go.mod
```diff
@@ -0,0 +1,14 @@
...
+go 1.13
+
+require (
+	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e
+	golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97
+	golang.org/x/net v0.0.0-20210716203947-853a461950ff
+	golang.org/x/oauth2 v0.0.0-20210628180205-a41e5a781914
+	gopkg.in/check.v1 v1.0.0-20201130134442-10cb98267c6c
+	gopkg.in/tomb.v2 v2.0.0-20161208151619-d5d1b5820637
+	gopkg.in/yaml.v2 v2.4.0
```
mvo5, line 12: Gustavo: we should move the entire project to v3 eventually

PR #134, opened 2021-09-08
Use greedy regex for $(HOST: )

zyga:
I cannot think of the reason the original code used non-greedy
matching, given that YAML allows us to control the text of the
string quite strongly, and there is nothing we want to parse after the
closing parenthesis.

This fixes variables which use ) inside the shell command.

### DIFF
--- a/spread/project.go
+++ b/spread/project.go
@@ -1090,8 +1090,8 @@ func (s stringer) String() string { return string(s) }
 
 var (
 	varname = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*(?:/[a-zA-Z0-9_]+(?:,[a-zA-Z0-9_]+)*)?$`)
-	varcmd  = regexp.MustCompile(`\$\(HOST:.+?\)`)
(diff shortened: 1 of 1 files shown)


mvo5:
<mvo> zyga-mbp: I quickly looked at the PR, if you could add an example how you noticed it and a "before/after" the change that would be lovely, then I can write a test for you
<zyga-mbp> yeah, I can even write the test but I didn't get to it yet, it's something I bumped into yesterday
<zyga-mbp> the example was $(HOST: $(echo example))
<zyga-mbp> or maybe
<zyga-mbp> the example was $(HOST: $(echo example)-foo)
<zyga-mbp> anyway, I'll update it with unit tests
<mvo> zyga-mbp: cool, that is all I need
<zyga-mbp> the real code was in libzt test suite, I've updated it to support local snap store proxy for caching
<zyga-mbp> and I've used quite a bit of HOST variables for that
<zyga-mbp> I can paste the real thing but the point is that the old spread stopped at the first right ")"
<zyga-mbp> so $(HOST: ...) could never have ) insdie
<zyga-mbp> *inside

Issue #81, opened 2019-07-25
Allow LXD profiles to be passed when launching


SimonRichardson:
When launching a series of tasks on LXD backends it would be ideal if we could also passing in a series of profile names to also be used when launching[1]. This would allow for things like nesting or privileged containers.

1. https://github.com/snapcore/spread/blob/c85b89fe2746a509b2b9f28be5c4dce1130d86f1/spread/lxd.go#L110

PR #136, opened 2021-10-26
lxd: support for container profiles

Meulengracht:
To improve the support for LXD containers the following PR adds the ability to provide profiles for container configurations. This allows containers to be configured for a specific purpose (like using privileged containers or device configurations etc).

This adds a new configuration parameter for the LXD backend that allows setting multiple profiles in a comma-seperated list.

This implements issue #81

### DIFF
--- a/spread/export_test.go
+++ b/spread/export_test.go
@@ -38,6 +38,20 @@ func FakeSshDial(f func(network, addr string, config *ssh.ClientConfig) (*ssh.Cl
 	}
 }
 
+type LXDServerJSON = lxdServerJSON
+
+func LXDProviderServerJSON(provider Provider, name string) (*LXDServerJSON, error) {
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -792,6 +792,15 @@ backends:
                 image: ubuntu:16.04.1
 ```
 
+Additionally, you can provide a comma-seperated list of LXD profiles to use:
+```
+backends:
+    lxd:
+        profiles: my-profile
```
anonymouse64, line 799: probably this setting should be more specifically named, maybe `container-profiles` ?

Meulengracht: This is a good point, it would avoid any ambigitity with other backends in the future.

anonymouse64 (approved):
seems fine to me, needs review from others however

PR #137, opened 2021-11-16
New option to select the number of disks on google instances

sergiocazzolato:
The idea is to have a new option to allow to define the number of disks
for a machine in google backend servers.
The disks number can be set either at backend or at system level.


mardy (approved):
LGTM, but don't we need to document it somewhere?

PR #138, opened 2021-11-16
New option to show the output for successful tasks

sergiocazzolato:
The idea is to display the output of a task even it is passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,8 @@ Spread
 [Rebooting](#rebooting)  
 [Timeouts](#timeouts)  
 [Fast iterations with reuse](#reuse)  
-[Debugging](#debugging)  
+[Debugging](#debugging)
+[Verbosity](#verbosity)
 [Repeating tasks](#repeating)
 [Passwords and usernames](#passwords)  
 [Including, excluding, and renaming files](#including)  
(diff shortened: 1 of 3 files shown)


jnsgruk:
Would love to see this merged!

One thing that strikes me is that we should probably use a different option name, given that `-vv` has been well established for some time now.

Perhaps the option could be renamed `-show-task-output` or similar?

niemeyer (changes requested):
This PR has been left unattended for years now. It has no motivation in its description, and its implementation is broken because it modifies the CLI behavior for options that have been out for a decade by now.

All I can do is guess why this was proposed, so some quick feedback:

1) One of the seed motivations for the creation of Spread back then was because autopkgtests offers a wall of text, and then you have to browse against a never-ending scrolling of pages to try to make sense of what it is saying. We should not bring that back.

2) You can get a list of which tasks are run, and thus a sense of "progress", by simply providing -v. I would discourage its use in CI because, again, you're forcing everyone to see a very long list of tasks that are irrelevant, instead of focusing on the things that do need attention.

3) If the goal is seeing things that are hanging, Spread does that automatically, with a configurable timeout (see warning-timeout), which will display the output until the point where it has stopped.

4) If the goal is to debug tasks that are failing, you can get a comprehensive output by using -vv today already.

So all of this are independent reasons why this could be unnecessary and unwanted. Ideally, though, such changes should be discussed in advance so that we can tell why something is being proposed at all, and what might be the best way of doing it.

PR #142, opened 2021-12-06
spread: fix bios path check, be more verbose about choosing the file

bboozzoo:
Fix BIOS path check, which called os.Stat() with incorrect path and thus never
used the bios file under ~/.spread/qemu/bios.


mvo5 (approved):
Thank you, this is very nice (and a bit embarrassing that I did not notice this when working on the feature).

anonymouse64 (approved):
lgtm

PR #149, opened 2022-05-06
Add xunit and jsonunit reports

sergiocazzolato:
The idea is to generate a report with the results including which tests were successfull, failed and aborted

This can be used to define which tests have to be re-executed


spread/jsonunit.go
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+	suite := NewJSONUnitTestSuite(suiteName)
+
+	tss.addSuite(suite)	
+	return suite
+}
+
+type JSONUnitTestSuite struct {
+	Successfull int              `json:"successfull,attr"`
```
mardy, line 30: there's one "l" too many: "successful"

spread/jsonunit.go
```diff
@@ -0,0 +1,168 @@
...
+	TestCases   []*JSONUnitTestCase
+}
+
+func NewJSONUnitTestSuite(suiteName string) *JSONUnitTestSuite {
+	return &JSONUnitTestSuite{
+		Successfull: 0,
+		Failed: 0,
+		Aborted: 0,
+		Total: 0,
+		Time: "",
```
mardy, line 45: You don't need these initializations, they are already the default ones.

mardy (commented):
I didn't review it, I stopped when I noticed that — if I'm not wrong — most of the code between the JSONUnit and the XUnit backend is nearly the same. Maybe you could write a backend taht fills a generic data structure, and then have it reused by the two backends, which would only differ in the serialization. I'm not sure if the structure of the two format is that similar, but notice that you can have both json and xml tags on the same structure field; so, if you are lucky, you don't even need to create different structures for the two backends.

Also please notice that the file is not formatter properly (some lines use spaces instead of tabs); you can use `gofmt` to fix them automatically.

spread/jsonunit.go
sergiocazzolato: nice, thanks

spread/jsonunit.go
sergiocazzolato: nice, thanks

spread/runner.go
```diff
@@ -36,6 +36,8 @@ type Options struct {
 	Seed           int64
 	Repeat         int
 	GarbageCollect bool
+	XUnit          bool
+	JSONUnit          bool
```
bboozzoo, line 40: indentation seems off here, did you run gofmt?

spread/runner.go
```diff
@@ -1073,6 +1081,44 @@ func (s *stats) log() {
 	logNames(printf, "Failed project restore", s.ProjectRestoreError, projectName)
 }
 
+func (s *stats) addTestsToReport(report Report, testsList []*Job, result string, verb string) {
+	for _, job := range testsList {
+		splittedName := strings.Split(taskName(job), "/")
+		suiteName := strings.Join(splittedName[:len(splittedName)-1], "/")
+		testName := splittedName[len(splittedName)-1]
```
bboozzoo, line 1088:
```suggestion
		split := strings.Split(taskName(job), "/")
		suiteName := strings.Join(split[:len(split)-1], "/")
		testName := split[len(split)-1]
```
looks like this could have been done using strings.LastIndex() or filepath.Dir/Base() if the test name always look like paths, maybe something like:
```
// task names look like paths (eg suite/foo/bar/job)
name := taskName(job)
suiteName := filepath.Dir(name)
testName := filepath.Base(name)
```

spread/jsonunit.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func NewJSONUnitTestCases() *JSONUnitTestCases {
+	return &JSONUnitTestCases{
+		Successfull: 0,
+		Failed: 0,
+		Aborted: 0,
+		Total: 0,
+		TestCases: []*JSONUnitTestCase{},
+	}
```
bboozzoo, line 24:
```suggestion
	return &JSONUnitTestCases{}
```

spread/xunit.go
```diff
@@ -0,0 +1,163 @@
...
+
+func NewXUnitTestSuite(suiteName string) *XUnitTestSuite {
+	return &XUnitTestSuite{
+		Successfull: 0,
+		Failed: 0,
+		Aborted: 0,
+		Total: 0,
+		Name: suiteName,
+		TestCases: []*XUnitTestCase{},
+	}
```
bboozzoo, line 48:
```suggestion
	return &XUnitTestSuite{}
```

spread/jsonunit.go
```diff
@@ -0,0 +1,136 @@
...
+	Backend     string            `json:"backend,attr"`
+	System      string            `json:"system,attr"`
+	Name        string            `json:"name,attr"`
+	Suite       string            `json:"suite,attr"`
+	FullName    string            `json:"fullname,attr"`
+	Details     []*JSONUnitDetail `json:"details,omitempty"`
+}
+
+func NewJSONUnitTestCase(testName string, backend string, system string, suiteName string) *JSONUnitTestCase {
+	fullname := backend + ":" + system + ":" + suiteName + "/" + testName
```
bboozzoo, line 64:
```suggestion
	fullname := fmt.Sprintf("%s:%s:%s/%s", backend, system, suiteName, testName)
```

spread/report.go
```diff
@@ -0,0 +1,14 @@
+package spread
+
+type Report interface {
```
bboozzoo, line 3: interface names are usually verbs if possible, so `Reporter` ? does it need to be exported?

spread/runner.go
```diff
@@ -1073,6 +1081,44 @@ func (s *stats) log() {
 	logNames(printf, "Failed project restore", s.ProjectRestoreError, projectName)
 }
 
+func (s *stats) addTestsToReport(report Report, testsList []*Job, result string, verb string) {
```
bboozzoo, line 1084:
```suggestion
func (s *stats) addTestsToReport(report Report, testsList []*Job, result string, stage string) {
```

spread/xunit.go
```diff
@@ -0,0 +1,163 @@
...
+    		if len(test.Details) > 0 {
+    			if test.Details[0].Type == aborted {
+    				ts.Aborted += 1
+    				ts.Failed -= 1
+    			} 
+    			t.Details = append(t.Details, test.Details[0])
+    		}
+    		return
+    	}
+	}
```
bboozzoo, line 63: formatting is off here

spread/xunit.go
```diff
@@ -0,0 +1,163 @@
...
+		Successfull: 0,
+		Failed: 0,
+		Aborted: 0,
+		Total: 0,
+		Name: suiteName,
+		TestCases: []*XUnitTestCase{},
+	}
+}
+
+func (ts *XUnitTestSuite) addTest(test *XUnitTestCase) {
```
bboozzoo, line 51: looks very much JSONUnitTestCases, maybe you could have something like genericUnitTestsReporter which then can be embedded in respective variant?

cmd/spread/main.go
```diff
@@ -34,6 +34,8 @@ var (
 	seed           = flag.Int64("seed", 0, "Seed for job order permutation")
 	repeat         = flag.Int("repeat", 0, "Number of times to repeat each task")
 	garbageCollect = flag.Bool("gc", false, "Garbage collect backend resources when possible")
+	xunit          = flag.Bool("xunit", false, "Create a XUnit report")
```
bboozzoo, line 37: Just a question, does it make sense to have both, or could we just go with xunit?

spread/runner.go
sergiocazzolato: added, thanks

spread/runner.go
sergiocazzolato: Makes sense, I'll do the change

spread/xunit.go
sergiocazzolato: updated

spread/jsonunit.go
sergiocazzolato: done

spread/runner.go
sergiocazzolato: done

cmd/spread/main.go
sergiocazzolato: well, this has to be discussed.
I created both because the idea of xunit is to use it to display in the actions page, and the jsonunit is to query

sergiocazzolato: Perhaps we could have just jsonunit and then convert to xunit format with a tool, so we don't need the xunit report.
@bboozzoo does it make sense?

spread/xunit.go
sergiocazzolato: udpated

spread/xunit.go
sergiocazzolato: done

sergiocazzolato:
A new pr with a simplified implementation has been added replacing the current one

PR #155, opened 2022-09-29
support ssh configuration logic through sshd_config.d on google backend

sergiocazzolato:
Some images (like ubuntu kinetic) override the ssh configuration by using the new `sshd_config.d` directory and no longer ship a `/etc/sshd_config` file.

This change makes sure these systems will have the needed shell configuration to run spread.

As the first value wins, we need to use `00-` for the configuration file to make sure this is the first file with that config

### DIFF
--- a/spread/google.go
+++ b/spread/google.go
@@ -154,7 +154,9 @@ const googleStartupScript = `
 echo root:%s | chpasswd
 
 sed -i 's/^\s*#\?\s*\(PermitRootLogin\|PasswordAuthentication\)\>.*/\1 yes/' /etc/ssh/sshd_config
-test -d /etc/ssh/sshd_config.d && echo -e 'PermitRootLogin=yes\nPasswordAuthentication=yes' > /etc/ssh/sshd_config.d/00-spread-settings.conf
+sed -i 's/^PermitRootLogin=/#PermitRootLogin=/g' /etc/ssh/sshd_config.d/* || true
+sed -i 's/^PasswordAuthentication=/#PasswordAuthentication=/g' /etc/ssh/sshd_config.d/* || true
+test -d /etc/ssh/sshd_config.d && echo -e 'PermitRootLogin=yes\nPasswordAuthentication=yes' > /etc/ssh/sshd_config.d/00-spread.conf
 
 pkill -o -HUP sshd || true
 
--- a/spread/google.go
+++ b/spread/google.go
@@ -154,6 +154,7 @@ const googleStartupScript = `
 echo root:%s | chpasswd
 
 sed -i 's/^\s*#\?\s*\(PermitRootLogin\|PasswordAuthentication\)\>.*/\1 yes/' /etc/ssh/sshd_config
+test -d /etc/ssh/sshd_config.d && echo -e 'PermitRootLogin=yes\nPasswordAuthentication=yes' > /etc/ssh/sshd_config.d/00-spread-settings.conf
 
 pkill -o -HUP sshd || true
 


mvo5 (approved):
Thank you for fixing spread with Ubuntu 22.10!

spread/google.go
```diff
@@ -154,6 +154,7 @@ const googleStartupScript = `
 echo root:%s | chpasswd
 
 sed -i 's/^\s*#\?\s*\(PermitRootLogin\|PasswordAuthentication\)\>.*/\1 yes/' /etc/ssh/sshd_config
+test -d /etc/ssh/sshd_config.d && echo -e 'PermitRootLogin=yes\nPasswordAuthentication=yes' > /etc/ssh/sshd_config.d/00-spread-settings.conf
```
mvo5, line 157: Juerg just pointed that we probably need the above tweak in `client.go:SetupRootAccess()` too and probably also in lxd.go ?

mardy:
Can this branch fix https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1991745 ?

niemeyer (changes requested):
Look reasonablle. Only two comments:

1) What happens if there are other entries in that directory which ist these options?  Is it last match wins, or first match wins?  Should we comment them out first?  How about this: if the directory exists, comment out the options in all files that exist there, and then add a new file with our options.

2) The -settings is redundant with the .conf, and given (1) above, I think we can name this nicely as just "spread.conf", assuming that's the pattern there.

sergiocazzolato:
Updated the change.

spread/google.go
```diff
@@ -154,6 +154,9 @@ const googleStartupScript = `
 echo root:%s | chpasswd
 
 sed -i 's/^\s*#\?\s*\(PermitRootLogin\|PasswordAuthentication\)\>.*/\1 yes/' /etc/ssh/sshd_config
+sed -i 's/^PermitRootLogin=/#PermitRootLogin=/g' /etc/ssh/sshd_config.d/* || true
+sed -i 's/^PasswordAuthentication=/#PasswordAuthentication=/g' /etc/ssh/sshd_config.d/* || true
+test -d /etc/ssh/sshd_config.d && echo -e 'PermitRootLogin=yes\nPasswordAuthentication=yes' > /etc/ssh/sshd_config.d/00-spread.conf
```
thp-canonical, line 159: According to the [`sshd_config(5)`](https://manpages.ubuntu.com/manpages/jammy/en/man5/sshd_config.5.html) manpage, the config file usually doesn't have a `=` between the key/value pairs. For example:

```
% grep = /etc/ssh/sshd_config
# This sshd was compiled with PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games
```

spread/google.go
```diff
@@ -154,6 +154,9 @@ const googleStartupScript = `
 echo root:%s | chpasswd
 
 sed -i 's/^\s*#\?\s*\(PermitRootLogin\|PasswordAuthentication\)\>.*/\1 yes/' /etc/ssh/sshd_config
+sed -i 's/^PermitRootLogin=/#PermitRootLogin=/g' /etc/ssh/sshd_config.d/* || true
+sed -i 's/^PasswordAuthentication=/#PasswordAuthentication=/g' /etc/ssh/sshd_config.d/* || true
```
ZeyadYasser, line 158: Config files might have whitespace before the the entry. also as @thp-canonical mentioned, the entry usually doesn't have a `=`.

```suggestion
sed -i 's/^\s*\(PermitRootLogin\|PasswordAuthentication\)\>.*/# COMMENTED OUT BY SPREAD: \0/' /etc/ssh/sshd_config.d/* || true
```

ZeyadYasser (changes requested):

ZeyadYasser:
@sergiocazzolato could you please also extend the fix to the LXD backend?

It might be a good idea to extend the LXD spread test to run with other system variants such as ubuntu-22.04.
https://github.com/snapcore/spread/blob/ded9133cdbceaf01f8a1c9decf6ff9ea56e194d6/tests/lxd/task.yaml#L12-L13
We could also extend the base spread.yaml in a follow up PR to add multiple systems most importantly 22.04 to test the google backend. https://github.com/snapcore/spread/blob/ded9133cdbceaf01f8a1c9decf6ff9ea56e194d6/spread.yaml#L17-L31

PR #159, opened 2023-02-06
client: use kernel `boot_id` to detect if a system got rebooted

mvo5:
This commit changes the code that detects if a reboot has happend based on the kernel `boot_id`. This is based on the idea from Valentin in PR#143 but slightly simplified as it will not parse the uuid but just tread as an opaque string.

### DIFF
--- a/spread/client.go
+++ b/spread/client.go
@@ -74,7 +74,7 @@ func (c *Client) ResetJob() {
 	c.SetJob("")
 }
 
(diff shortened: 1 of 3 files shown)


niemeyer (approved):
That's nice, thank you. It's also nice that the return of the function is an opaque string, as we can easily change the implementation again in the future if required.

PR #166, opened 2023-04-12
qemu: allow specifying the machine type

IsaacJT:
This allows QEMU machine types to be used other than the default "pc", such as "q35". Valid alternative machine types can be identified with the `qemu-system-x86_64 -M help` command.

This is useful for testing images for embedded systems which may not support the "pc" QEMU machine type, or require features from "q35" or other types.

### DIFF
--- a/spread/project.go
+++ b/spread/project.go
@@ -139,6 +139,10 @@ type System struct {
 
 	Priority OptionalInt
 	Manual   bool
+
+	// Type of machine emulated, only for qemu. Defaults to "pc", which is
+	// the QEMU default for x86_64
+	MachineType string `yaml:"machine-type"`
 }
 
 func (system *System) String() string { return system.Backend + ":" + system.Name }
--- a/spread/qemu.go
+++ b/spread/qemu.go
@@ -121,6 +121,13 @@ func biosPath(biosName string) (string, error) {
 	return "", fmt.Errorf("cannot find bios path for %q", biosName)
 }
 
+func machineType(system *System) string {
+	if system.MachineType == "" {
+		return "pc"
+	}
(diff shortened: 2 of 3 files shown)


spread/project.go
```diff
@@ -139,6 +139,10 @@ type System struct {
 
 	Priority OptionalInt
 	Manual   bool
+
+	// Type of machine emulated, only for qemu. Defaults to "pc", which is
+	// the QEMU default for x86_64
+	MachineType string `yaml:"machine-type"`
```
mvo5, line 145: We try to avoid backend specific options in System, if possible we try to find a generic term that could be reused across various backends. MachineType (or just `Machine`) is actually not bad as it potentially  maps to many different backends, e.g. in linode or google it could map to the machine type like: `e2-standard-2`. Three is an existing `Plan` in project.go that is currently used for this.

Which makes me wonder if could consolidate `plan` and `machine` somehow, I see some optons:
1. add new `machine`  and only use in qemu
2. use `plan` in the qemu backend and map that to machine type
3. add new `machine`  and use it in linode/google instead of `plan` (deprecate `plan`)
4. ... something I forgot?

Each has different tradeoffs so input from @niemeyer on this would be best.

IsaacJT: Thanks @mvo5, modifying the code to use the existing `plan` variable sounds like a good idea to me as it's currently not in use by the `qemu` backend. If you and @niemeyer are happy with that I will adjust the PR

IsaacJT: Implemented here: https://github.com/snapcore/spread/pull/169

I'll keep this PR open in case we want to go in a different direction

PR #167, opened 2023-04-12
qemu: add options to enable virtio device backends

IsaacJT:
In some cases, the virtio backend is needed for the disk. This adds the boolean option "virtio-disk", which specifies if the disk should be attached via the virtio backend when true, or using the default backend when false.

Similarly, in some cases the virtio backend is needed for the network. This also adds the option "virtio-net", which specifies if the network interface should be attached via the virtio backend when true, or using the default e1000 backend when false.

This is useful for testing images with stripped-down kernels which do not provide drivers for the default network and storage backends, and only support virtio.

### DIFF
--- a/spread/project.go
+++ b/spread/project.go
@@ -143,6 +143,10 @@ type System struct {
 	// Specify that the disk should be attached using the "virtio" backend
 	// in QEMU. Only supported by the qemu backend.
 	VirtioDisk bool `default:"false" yaml:"virtio-disk"`
+
+	// Specify that the network interface should be attached using the
+	// "virtio" backend in QEMU. Only supported by the qemu backend.
+	VirtioNet bool `default:"false" yaml:"virtio-net"`
 }
 
 func (system *System) String() string { return system.Backend + ":" + system.Name }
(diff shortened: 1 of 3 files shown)


spread/project.go
```diff
@@ -139,6 +139,14 @@ type System struct {
 
 	Priority OptionalInt
 	Manual   bool
+
+	// Specify that the disk should be attached using the "virtio" backend
+	// in QEMU. Only supported by the qemu backend.
+	VirtioDisk bool `default:"false" yaml:"virtio-disk"`
```
mvo5, line 145: This is fairly specifc to qemu right now, but the `System` struct is used across all backends. So a more generic mechanism sounds more useful, so maybe a string option like: `disk-driver` or `virtual-disk-driver` or similar and then the user could set it to `virtio-disk`  on qemu and on different backends there could be a different value here (same for the network of course).

I also wonder if this could become a map even, so it would be something like:
```
virtual-drivers:
 disk: virtio-disk
 network: virtio-net
```
and depending on the backend it would be interpreted differently.

IsaacJT: Great suggestion. Should this fail on systems that do not implement this, or should they just ignore the options that are specified?

PR #168, opened 2023-05-02
spread: add device backend configuration support to QEMU

IsaacJT:
This adds the `device-backends` configuration option, which allows the user  to specify which backends for specific devices the system should use. This has no effect on its own, as the implementation is system backend specific; only the QEMU backend has been enabled with this functionality.

For the QEMU backend, this allows the user to specify which backend should be used for the disk and network devices. Currently, only the "network" and "disk" devices are able to be specified. If they are not specified, QEMU will use the default backends ("e1000" and "none", respectively).

This is useful for testing images with stripped-down kernels which do not provide drivers for the default network and storage backends, and only support other drivers (for example only virtio).

### DIFF
--- a/README.md
+++ b/README.md
@@ -852,6 +852,30 @@ adt-buildvm-ubuntu-cloud
 ```
 When done move the downloaded image into the location described above.
 
+The QEMU backend will normally create the VM with the default device
+backends. The backends for the network and drive devices can be overriden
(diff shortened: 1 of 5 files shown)


IsaacJT:
This is the more generic iteration of my previous PR https://github.com/snapcore/spread/pull/167

PR #170, opened 2023-06-02
openstack backend

sergiocazzolato:
This is the new openstack backend. Through this backend it is possible to run tests/tasks in openstack infrastructure. 

The documenatation is also added explaning how to setup spread to use it.

For the openstack backend implementation the lib goose was used. This lib provides the clients needed to interact with the different openstack modules (nova. neutron, glance and, keystone).


spread/openstack.go
```diff
@@ -0,0 +1,565 @@
...
+		FlavorId:         flavor.Id,
+		ImageId:          image.Id,
+		AvailabilityZone: availabilityZone.Name,
+		Networks:         []nova.ServerNetworks{{NetworkId: network.Id}},
+		Metadata:         tags,
+		UserData:         []byte(cloudconfig),
+	}
+	server, err := p.computeClient.RunServer(opts)
+	if err != nil {
+		return nil, &FatalError{fmt.Errorf("Could not create instance", err)}
```
corytodd, line 398: Should the format string include a format specifier here?

corytodd:
Awesome, we are able to allocate and discard on our openstack tenant. Do you think it's feasible to support adding security groups automatically? For example, we do not use a default `allow-ssh` policy on our network so in order for these tests to work I had to manually attach the policy after allocation but before the test ran. We have other ports that may need to be open depending on the test so having a way to set these dynamically would be useful for us.

sergiocazzolato:
> Awesome, we are able to allocate and discard on our openstack tenant. Do you think it's feasible to support adding security groups automatically? For example, we do not use a default `allow-ssh` policy on our network so in order for these tests to work I had to manually attach the policy after allocation but before the test ran. We have other ports that may need to be open depending on the test so having a way to set these dynamically would be useful for us.

Thanks for taking a look. I'll add few extra features:
 add extra storage
 specify the network
 specify the security group

spread/openstack.go
sergiocazzolato: not needed, thanks

sergiocazzolato: the FatalError means it is not needed to retry and it is built from an error, I already updated the error in the backend to make sure we retry only when it makes sense

corytodd: Oh okay, I see know thanks for explaining.

corytodd:
Specifying the network and security group work great! This passes our testing, I would support a +1 on getting this merged. Thanks!

spread/openstack.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+	return findErrorTitle(node)
+}
+
+func (p *openstackProvider) findFlavor(flavorName string) (nova.Entity, error) {
+	var flavor nova.Entity
+
+	flavors, err := p.computeClient.ListFlavors()
+	if err != nil {
+		return flavor, fmt.Errorf("failed to retrieve flavors list: %s", errorTitle(err.Error()))
```
mvo5, line 214: (nitpick) this should be `cannot retrieve...` to be consistent with the way errors are reported

spread/openstack.go
```diff
@@ -0,0 +1,631 @@
...
+
+	for _, f := range flavors {
+		if f.Name == flavorName {
+			flavor = f
+			break
+		}
+	}
+
+	if flavor.Id == "" {
+		return flavor, fmt.Errorf("specified flavor not found: %s", flavorName)
```
mvo5, line 225: (nitpick) same as above, for consistency this should be `cannot find flavor %s` (and bunch more in the error handling)

spread/openstack.go
```diff
@@ -0,0 +1,631 @@
...
+// which contain title and details related to the error
+func errorTitle(msg string) string {
+	node, err := html.Parse(strings.NewReader(msg))
+	if err != nil {
+		return ""
+	}
+	return findErrorTitle(node)
+}
+
+func (p *openstackProvider) findFlavor(flavorName string) (nova.Entity, error) {
```
mvo5, line 209: (nitpick) it seems to me that using a pointer to nova.Entity here would be slightly better as you can return `nil` instead of an "empty" nova.Entity. Plus when found just returning the value instead of break and indirect return is probably slightly simpler. 

I.e.:
```go
diff --git a/spread/openstack.go b/spread/openstack.go
index 52c6aaf..417e67a 100644
--- a/spread/openstack.go
+++ b/spread/openstack.go
@@ -206,26 +206,19 @@ func errorTitle(msg string) string {
        return findErrorTitle(node)
 }
 
-func (p *openstackProvider) findFlavor(flavorName string) (nova.Entity, error) {
-       var flavor nova.Entity
-
+func (p *openstackProvider) findFlavor(flavorName string) (*nova.Entity, error) {
        flavors, err := p.computeClient.ListFlavors()
        if err != nil {
-               return flavor, fmt.Errorf("failed to retrieve flavors list: %s", errorTitle(err.Error()))
+               return nil, fmt.Errorf("failed to retrieve flavors list: %s", errorTitle(err.Error()))
        }
 
        for _, f := range flavors {
                if f.Name == flavorName {
-                       flavor = f
-                       break
+                       return &f, nil
                }
        }
 
-       if flavor.Id == "" {
-               return flavor, fmt.Errorf("specified flavor not found: %s", flavorName)
-       }
-
-       return flavor, nil
+       return nil, fmt.Errorf("specified flavor not found: %s", flavorName)
 }
 
 func (p *openstackProvider) findNetwork(name string) (neutron.NetworkV2, error) {
```

spread/openstack.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+
+	if flavor.Id == "" {
+		return flavor, fmt.Errorf("specified flavor not found: %s", flavorName)
+	}
+
+	return flavor, nil
+}
+
+func (p *openstackProvider) findNetwork(name string) (neutron.NetworkV2, error) {
```
mvo5, line 231: (nitpick) same idea about pointer vs value return as above

spread/openstack.go
```diff
@@ -0,0 +1,631 @@
...
+			return title
+		}
+	}
+
+	return ""
+}
+
+// error messages returned by openstack api are html
+// which contain title and details related to the error
+func errorTitle(msg string) string {
```
mvo5, line 201: This is curious, I looked at the API in https://github.com/go-goose/goose/blob/abaee9096e3a/nova/nova.go#L130 and it would be nice too what error type you actually get here from nova. As a quick experiment `fmt.Sprintf("%v %T\n", err, err) ` would be great. I'm puzzled that this is needed.

spread/openstack.go
```diff
@@ -0,0 +1,631 @@
...
+	var sameImage glance.ImageDetail
+	var lastImage glance.ImageDetail
+	var lastCreatedDate time.Time
+
+	images, err := p.imageClient.ListImagesDetail()
+	if err != nil {
+		return sameImage, fmt.Errorf("failed to retrieve images list: %s", errorTitle(err.Error()))
+	}
+
+	for _, i := range images {
```
mvo5, line 259: I wonder if this code should be slightly more elaborate and follow `googleProvider:image` or `linode:tempate()`. linode is simpler and just does a prefix search but afaict all do more than just check for "contains"

spread/openstack.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+
+	if lastImage.Id != "" {
+		return lastImage, nil
+	}
+
+	return sameImage, &FatalError{fmt.Errorf("No matching image found")}
+}
+
+func (p *openstackProvider) findAvailabilityZone() (nova.AvailabilityZone, error) {
```
mvo5, line 290: same comemnt about pointer as above and "failed"->"cannot"

spread/openstack.go
```diff
@@ -0,0 +1,631 @@
...
+	var zone nova.AvailabilityZone
+
+	zones, err := p.computeClient.ListAvailabilityZones()
+	if err != nil {
+		return zone, fmt.Errorf("failed to retrieve availability zones: %s", errorTitle(err.Error()))
+	}
+
+	if len(zones) == 0 {
```
mvo5, line 298: This can be written as:
```go
	if len(zones) == 0 {
		return zone, &FatalError{errors.New("cannot find any availability zones")}
	}
	return zones[0], nil
}
```

spread/openstack.go
```diff
@@ -0,0 +1,631 @@
...
+	"github.com/go-goose/goose/v5/glance"
+	"github.com/go-goose/goose/v5/identity"
+	"github.com/go-goose/goose/v5/neutron"
+	"github.com/go-goose/goose/v5/nova"
+
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/net/context"
+	"golang.org/x/net/html"
+
+	"strconv"
```
mvo5, line 19: (nitpick) this should be part of the first block with the go provided libs.

mvo5 (commented):
Thanks for working on this Sergio! I did a first review and have some suggestion inline. I will also sync with Gustavo to ask how he wants to see this moving forward.

mvo5:
Some feedback I got from Gustavo:
- the names in "systems": "network" and "security-groups" should be looked at again, anything in this should be a generic term/concept that can be potentially reused in all the backends
- the way images are picked should be similar to google (or linode)
I tweaked the other things that came up during the review or that I noticed myself now :)

spread/project.go
```diff
@@ -64,6 +64,10 @@ type Backend struct {
 	Location string
 	Storage  Size
 
+	// Only for Openstack so far
+	Network        string
+	SecurityGroups []string `yaml:"security-groups"`
```
mvo5, line 69: Just as a datapoint, this can be any string, we have seen things like `juju-054c6083-a7ba-4e30-9c51-c2c58c74207d`  or `default`.

spread/project.go
```diff
@@ -64,6 +64,10 @@ type Backend struct {
 	Location string
 	Storage  Size
 
+	// Only for Openstack so far
+	Network        string
```
mvo5, line 68: This can be any string like: `network_production-router1` or `network_misc-some-juju-env`

mvo5 (commented):
I see two broad options for the question of how t deal with network/security-groups:
1. make it an optional and backend specific part of location: `snapd-spread/us-east1-b:some-network:some-security-group,some-security-group2
2. add a new generic "security:" (exact name TBD) tag that has both tags: "security: some-netwrok:some-security-group1,group2"

spread/project.go
mvo5: The feedback from Gustavo here is that this needs to be a slice []string because people may attach machines to multiple networks (in the general case)

spread/project.go
mvo5: The feedback from Gustavo for this one is that the more generic term `groups:` should be used here and that is then applied in openstack to security-groups and in GCE to the equivalent concept (VPC?)

mvo5 (commented):
Gustavo also asked that backends that do not (yet) support the options network/groups should error when they are specified. 

A smoke spread test against a real system should be included and unit tests as far as possible without modifying non-openstack code. 

The way images are selected/filtered also needs a review.

sergiocazzolato:
@mvo5 about the network list associated to a machine, the consideration here is that where there are more than 1 network associated to a machine, the ip used by spread to connect has to be provided by the first network. I'll include that in the README.

spread/openstack.go
sergiocazzolato: fmt.Sprintf("%v, err)
2023-07-17 16:26:39 authentication failed
caused by: requesting token: Unauthorised URL https://keystone.bos01.canonistack.canonical.com:5000/v3/auth/tokens
caused by: request (https://keystone.bos01.canonistack.canonical.com:5000/v3/auth/tokens) returned unexpected status: 401; error info: Failed: 401 error: The request you have made requires authentication.

fmt.Sprintf("%T, err)
2023-07-17 16:26:39 *errors.gooseError

sergiocazzolato: I could reproduce it in checkKey function

README.md
```diff
@@ -899,6 +900,61 @@ and need to be removed by hand.
...
+OS_INTERFACE
+OS_IDENTITY_API_VERSION
+```
+
+You can set up those variables by sourcing the Openstack RC file. 
+For more information about the environment setup please read 
+https://docs.openstack.org/ocata/user-guide/common/cli-set-environment-variables-using-openstack-rc.html
+
+Images are located by first attempting to match the provided value exactly
+against the image name, if there is not exact match the most recent image
```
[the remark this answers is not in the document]
sergiocazzolato: thanks for the review @mvo5. In gce the images have associated a name, a family of images and a project, but in openstack the images just have a name. So it is not possible to follow the same mechanism that is used in gce. 

examples:
 openstack: https://paste.ubuntu.com/p/GjHjJZGcyH/
 gce: https://paste.ubuntu.com/p/k4npC2H4kF/

spread/openstack.go
sergiocazzolato: @mvo5 could you please elaborate a bit more this? In openstack the images dont have family or project associated as in gce, so because of that I used the contain.

mvo5: Sorry, I was mostly wondering what contraints there are about image names, I created a unit test for the code now so that we can explore various test cases and examples :)

mvo5 (commented):
Thanks, I looekd a bit more and I really like the updated spread test! I also added a few comments and suggestions and pushed some small tweaks.

spread/openstack.go
```diff
@@ -0,0 +1,673 @@
...
+}
+
+type openstackImagesCache struct {
+	mu     sync.Mutex
+	ready  bool
+	images []openstackImage
+	err    error
+}
+
+func openstackName() string {
```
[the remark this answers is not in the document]
sergiocazzolato: Thanks for the review, so the name contains "month day time - microseconds"
This is the same approach we use for google instances.

This is an example: 
Name aug171434-560581
Creation time Aug 17, 2023, 11:34:36 AM UTC-03:00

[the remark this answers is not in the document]
mvo5: Aha, silly me, I forgot this is exactly the same as in the google backend. I reverted my change and improved the unit test a little bit to make sure it captures the right things.

spread/openstack.go
```diff
@@ -0,0 +1,679 @@
...
+#cloud-config
+runcmd:
+  - echo root:%s | chpasswd
+  - sed -i 's/^\s*#\?\s*\(PermitRootLogin\|PasswordAuthentication\)\>.*/\1 yes/' /etc/ssh/sshd_config
+  - test -d /etc/ssh/sshd_config.d && echo 'PermitRootLogin=yes' > /etc/ssh/sshd_config.d/00-spread.conf
+  - test -d /etc/ssh/sshd_config.d && echo 'PasswordAuthentication=yes' >> /etc/ssh/sshd_config.d/00-spread.conf
+  - pkill -o -HUP sshd || true
+`
+const openstackReadyMarker = "MACHINE-IS-READY"
+const openstackNameLayout = "Jan021504.000000"
```
mvo5, line 167: Fwiw, this and the openstackName() helper are the same as in google.go (googleName()) so we could extract a common helper. But maybe as a followup to make the review for this PR easier, one nice aspect about the current PR is that openstack.go and google.go are similar so the patterns are similar etc.

spread/openstack.go
```diff
@@ -0,0 +1,673 @@
...
+		if strings.HasPrefix(line, "caused by:") {
+			errorCause := strings.Split(line, "; error info:")[0]
+			return strings.TrimSpace(strings.TrimPrefix(errorCause, "caused by:"))
+		}
+	}
+	return ""
+}
+
+func errorMsg(osErr error) string {
+	msg := osErr.Error()
```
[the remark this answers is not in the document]
sergiocazzolato: There are examples of errors with details in html:
https://paste.ubuntu.com/p/3VMpmWjYYm/
https://paste.ubuntu.com/p/mJGVw72pSb/

I updated the code to show in the output the cause of the error and not the html title (included in the error info).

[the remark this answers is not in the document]
mvo5: Thanks, this is very useful, I am creating unit tests based on this now to help us decide what we want to do with the errors. One thought that occured to me while doing this is that "errorMsg" is probably too generic as it's a name available inside all of the "package spread" namespace.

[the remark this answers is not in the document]
mvo5: I guess we should talk to Gustavo about what he wants here, the current approach seems sensible but I can see an argument for simply dumping the whole error (with html/json) as it *may* contain useful information. But it certainly also contains a lot of waste that makes the errors harder to read.

README.md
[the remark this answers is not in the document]
mvo5: Can I get a listing of what the "openstack image list" for the spread environment that we use for testing looks like? My worry is a bit that right now the code says: `strings.Contains(i.Name, imageName)` which feels a bit too lose for my taste, I would love to make it a bit stricter by having common prefixes/suffies or some regexp like way to be more certain that the image selected is actually the right image :)

README.md
```diff
@@ -899,6 +900,61 @@ and need to be removed by hand.
...
+_$PROJECT/spread.yaml_
+```
+(...)
+
+backends:
+    openstack:
+        plan: cpu2-ram4-disk10
+        halt-timeout: 2h
+        systems:
+            - ubuntu-20.04:
```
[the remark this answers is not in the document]
sergiocazzolato: added

README.md
```diff
@@ -856,14 +857,14 @@ _$PROJECT/spread.yaml_
 backends:
     google:
         key: $(HOST:echo $GOOGLE_JSON_FILENAME)
-	location: yourproject/southamerica-east1-a
+    location: yourproject/southamerica-east1-a
```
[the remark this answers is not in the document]
sergiocazzolato: I think it was gofmt, should I revert it?

README.md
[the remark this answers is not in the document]
sergiocazzolato: Hi, thanks for checking this one again.

ps5: https://paste.ubuntu.com/p/4vfF3JJwQ9/
canonistack: https://paste.ubuntu.com/p/pwmNVXXGsC/

Just consider that these images lists follow the convention defined by the owner, but other openstack providers could use a different naming convention. So, if in a different environment the images name don't follow the pattern prefixes/suffies, it will be impossible to boot those images.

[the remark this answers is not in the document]
mvo5: Thanks for these examples. So it seems like (in general) we cannot force the openstack instance to follow a convention we prescribe. If so, maybe we can be more precise. It seems there are two general cases:
1. The name is slightly weird/long like `auto-sync/ubuntu-bionic-18.04-amd64.img`. This should be fine and the user just needs to jut the long name in.
2. The name requires sorting (like `ubuntu/ubuntu-bionic-18.04-amd64-server-20220302-disk1.img`). But given that each image has a "created" property we can simply sort by that and use the latest image. So a strings.Contains() seems okay here. We could require a regex instead but it seems for the examples the `strings.Contains` plus sorting by creation date is sufficient. Wdyt @niemeyer ?

spread/openstack_test.go
```diff
@@ -0,0 +1,153 @@
...
+<h1>Service Unavailable</h1>
+<p>The server is temporarily unable to service your
+request due to maintenance downtime or capacity
+problems. Please try again later.</p>
+<hr>
+<address>Apache/2.4.29 (Ubuntu) Server at 10.48.7.10 Port 5000</address>
+</body></html>`)
+
+func (s *openstackSuite) TestErrorMsg(c *C) {
+	msg1 := spread.ErrorMsg(opstErr1)
```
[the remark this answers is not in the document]
mvo5: Gustavo: only have full output when enabling debug output!

spread/openstack.go
```diff
@@ -0,0 +1,658 @@
...
+const openstackCloudInitScript = `
+#cloud-config
+runcmd:
+  - echo root:%s | chpasswd
+  - sed -i 's/^\s*#\?\s*\(PermitRootLogin\|PasswordAuthentication\)\>.*/\1 yes/' /etc/ssh/sshd_config
+  - test -d /etc/ssh/sshd_config.d && echo 'PermitRootLogin=yes' > /etc/ssh/sshd_config.d/00-spread.conf
+  - test -d /etc/ssh/sshd_config.d && echo 'PasswordAuthentication=yes' >> /etc/ssh/sshd_config.d/00-spread.conf
+  - pkill -o -HUP sshd || true
+`
+const openstackReadyMarker = "MACHINE-IS-READY"
```
mvo5, line 165: This is currently unused, is there a way to have an echo on the console so that the machine-ready detection works similarly to google?

sergiocazzolato: Removed. 
In google we can use the api to read the serial console of the created instance, but the api client we use in openstack doesn't have that feature.

spread/openstack_test.go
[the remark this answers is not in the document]
sergiocazzolato: well, we are already returning the first "caused by" line

README.md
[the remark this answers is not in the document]
mvo5: This is done now, the image selection is now following what google does as much as possible, i.e. exact match wins always and if not the same "term" based matching as google is used. It's a bit more fuzzy/flexible compared to substring matching, I added a test to show this. I think this is what Gustavo expects but I will double check with him to be certain :)

spread/openstack_test.go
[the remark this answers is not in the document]
mvo5: I updated the code a bit now to follow Gustavos advise. Fwiw, the existing code was already using the last "caused by" because the for() loop did not exiting on the first match so it was already picking the last :) I also added a comment that we need to see if how well the error parsing works in practice (only real world errors will tell). I also added that the full error is always show when debug is enabled.

spread/openstack.go
mvo5: I added a TODO for now and will check with Gustavo if he feels strongly about this. It seems a bit unclear if it will work on any openstack instance.

spread/openstack.go
```diff
@@ -0,0 +1,640 @@
...
+	// The timeout for this process is 180 seconds
+	err = p.waitServerCompleteBuilding(s, 120)
+	if err != nil {
+		if p.removeMachine(ctx, s) != nil {
+			return nil, &FatalError{fmt.Errorf("cannot allocate or deallocate (!) new openstack server %s: %v", s, err)}
+		}
+		return nil, &FatalError{fmt.Errorf("cannot allocate new openstack server %s: %v", s, err)}
+	}
+
+	// Connect through ssh to the
```
ZeyadYasser, line 508:
```suggestion
	// Connect through ssh to the instance
```

spread/openstack.go
```diff
@@ -0,0 +1,640 @@
...
+			Status:   openstackProvisioning,
+			Created:  time.Now(),
+		},
+
+		system: system,
+	}
+
+	// First we need to wait until the image is active and there is no erros during the spawning process
+	// The timeout for this process is 180 seconds
+	err = p.waitServerCompleteBuilding(s, 120)
```
ZeyadYasser, line 500: minor nitpick, in the comment it says 180s but 120s is being passed.

spread/openstack.go
```diff
@@ -0,0 +1,640 @@
...
+		secGroupNames = append(secGroupNames, nova.SecurityGroupName{Name: name})
+	}
+	return secGroupNames, nil
+}
+
+func (p *openstackProvider) waitServerCompleteBuilding(s *openstackServer, timeoutSeconds int) error {
+	// Wait until the server is actually running
+	start := time.Now()
+	for {
+		if time.Since(start) > time.Duration(timeoutSeconds)*time.Second {
```
ZeyadYasser, line 344: would it be a better idea to use channel timeouts to follow waiting pattern done in `googleProvider.waitOperation`?

spread/openstack.go
```diff
@@ -0,0 +1,640 @@
...
+	addr := s.address
+	if !strings.Contains(addr, ":") {
+		addr += ":22"
+	}
+
+	// Iterate until the ssh connection to the host can be stablished
+	// In openstack the client cannot access to the serial console of the instance
+	start := time.Now()
+	for {
+		if time.Since(start) > time.Duration(timeoutSeconds)*time.Second {
```
ZeyadYasser, line 394: same comment as above regarding following timeout pattern in `googleProvider.waitOperation`.

ZeyadYasser (commented):
Great work, just tiny comments and nitpicks.

spread/openstack.go
mvo5: Yes, that would probably a nicer pattern here, 

Here is an (untested) diff, I wanted to write a proper unit test to ensure this works correctly but it's too late in the evening for it :/
[waitOpPattern.diff.txt](https://github.com/snapcore/spread/files/12921775/waitOpPattern.diff.txt)

spread/openstack.go
```diff
@@ -0,0 +1,641 @@
...
+		return nil, fmt.Errorf("cannot retrieve flavors list: %s", errorMsg(err))
+	}
+
+	for _, f := range flavors {
+		if f.Name == flavorName {
+			return &f, nil
+		}
+	}
+
+	return nil, &FatalError{fmt.Errorf("cannot find valid flavor with name %s", flavorName)}
```
mvo5, line 214: This should be a `%q` instead of a %s

spread/openstack.go
```diff
@@ -0,0 +1,641 @@
...
+				lastCreatedDate = currCreatedDate
+			}
+		}
+	}
+
+	if lastImage.Id != "" {
+		return &lastImage, nil
+	}
+
+	return nil, &FatalError{fmt.Errorf("cannot find matching image for %s", imageName)}
```
mvo5, line 297: This should also be a `%q`

spread/openstack.go
```diff
@@ -0,0 +1,641 @@
...
+	for _, name := range names {
+		found := false
+		for _, sg := range secGroups {
+			if name == sg.Name {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return nil, &FatalError{fmt.Errorf("cannot find valid group with name %s", name)}
```
mvo5, line 333: Also here `%q`

spread/openstack.go
```diff
@@ -0,0 +1,758 @@
...
+
+func (p *openstackProvider) waitServerBootSerial(ctx context.Context, s *openstackServer) error {
+	timeout := time.After(openstackServerBootTimeout)
+	relog := time.NewTicker(60 * time.Second)
+	defer relog.Stop()
+	retry := time.NewTicker(openstackServerBootRetry)
+	defer retry.Stop()
+
+	var marker = openstackReadyMarker
+	for {
```
ZeyadYasser, line 451: @sergiocazzolato  do we need this loop? does the serial console take some time to be ready?

spread/openstack.go
```diff
@@ -0,0 +1,758 @@
...
+	}
+	return nil
+}
+
+func (p *openstackProvider) address(ctx context.Context, s *openstackServer) (string, error) {
+	timeout := time.After(30 * time.Second)
+	retry := time.NewTicker(5 * time.Second)
+	defer retry.Stop()
+
+	for {
```
ZeyadYasser, line 498: @sergiocazzolato do we need this loop? does the IP take some time to be ready after the server is provisioned?

spread/openstack.go
```diff
@@ -0,0 +1,758 @@
...
+	}
+
+	err = p.waitProvision(ctx, s)
+	if err == nil {
+		s.address, err = p.address(ctx, s)
+	}
+	if err == nil {
+		err = p.waitServerBoot(ctx, s)
+	}
+	// TODO: do we need to drop cloud-init script?
```
ZeyadYasser, line 625: @sergiocazzolato do we need to drop cloud-init script?

ZeyadYasser (commented):
Thank you, I have some small questions.

spread/openstack.go
sergiocazzolato: yes, the address takes time to be associated to the instace, usually the instance is allocated and after few iterations it has an ip associated to its network interface

spread/openstack.go
ZeyadYasser: sorry my bad, I meant in `getSerialConsoleOutput`, it was a loop before, I updated it.

spread/openstack.go
sergiocazzolato: no, we don't because in this script we configure ssh to allow the root connection using password and also it displays the marker that indicates the machine is ready

ZeyadYasser: Those should run in the first boot only, right?

README.md
```diff
@@ -899,6 +900,66 @@ and need to be removed by hand.
...
+OS_PROJECT_NAME
+OS_USER_DOMAIN_NAME
+OS_PROJECT_DOMAIN_ID (optional)
+OS_TENANT_ID
+OS_TENANT_NAME
+OS_USERNAME
+OS_PASSWORD
+OS_REGION_NAME
+OS_INTERFACE
+OS_IDENTITY_API_VERSION
```
niemeyer, line 948: In addition to the above, which of those variables have a direct equivalent in the Google backend setup?

spread.yaml
```diff
@@ -31,9 +36,13 @@ backends:
                 password: ubuntu
 
 exclude:
+    - .git
     - .spread-reuse.yaml
     - tests/.spread-reuse.yaml
     - $CACHE_DISABLED
+    - "*.snap"
```
niemeyer, line 43: Why the quotes only on this one?  Also, where do the snap files come from?

spread/export_openstack_test.go
```diff
@@ -0,0 +1,89 @@
+package spread
+
+import (
+	"context"
+	"time"
+
+	gooseClient "github.com/go-goose/goose/v5/client"
```
niemeyer, line 7: s/gooseClient/gooseclient/

Package names in Go are not typically cammel-cased.

spread/export_openstack_test.go
```diff
@@ -0,0 +1,89 @@
...
+	OpenstackName = openstackName
+)
+
+type (
+	OpenstackProvider   = openstackProvider
+	GlanceImageClient   = glanceImageClient
+	OpenstackServerData = openstackServerData
+)
+
+func MockOpenstackImageClient(opst *OpenstackProvider, newIC glanceImageClient) (restore func()) {
```
niemeyer, line 21: newIC, newIC, but then "timeout" and "retry" shortly below.  Let's stick to the latter convention please and have "imageClient" and "computeClient" instead as these are more readable.

spread/project.go
```diff
@@ -529,7 +537,7 @@ func Load(path string) (*Project, error) {
 			backend.Type = bname
 		}
 		switch backend.Type {
-		case "google", "linode", "lxd", "qemu", "adhoc", "humbox":
+		case "google", "linode", "lxd", "qemu", "adhoc", "humbox", "openstack":
```
niemeyer, line 540: Just for aesthetics, can we please move it after google?  It's the best analogous in that list.

tests/openstack/spread.yaml
```diff
@@ -0,0 +1,50 @@
...
+                  groups:
+                        - default
+                  networks:
+                        - private
+            - cirros-64-wrong-network:
+                  image: cirros-0.6.2-x86_64-disk
+                  groups:
+                        - default
+                  networks:
+                        - noexist
```
niemeyer, line 19: "noexist" no exist in English. :smile:

"invalid" works better and is more clear. There are more occurrences.

tests/openstack/task.yaml
```diff
@@ -0,0 +1,101 @@
...
+kill-timeout: 30m
+
+prepare: |
+    if [ ! -f .spread-reuse.yaml ]; then
+        touch /run/spread-reuse.yaml
+        ln -s /run/spread-reuse.yaml .spread-reuse.yaml
+    fi
+
+    sudo -u stack -i /opt/stack/devstack/unstack.sh
+    sudo -u stack -i /opt/stack/devstack/stack.sh
```
niemeyer, line 14: Where are these coming from?

tests/openstack/task.yaml
```diff
@@ -0,0 +1,101 @@
...
+            break
+        fi
+        sleep 1
+    done
+    test -z "$(openstack server list)"
+
+    # Check the error in case the network does not exist
+    spread openstack:cirros-64-wrong-network: -v -reuse -resend &> task.out || true
+    grep 'cannot find valid network with name "noexist"' task.out
+    test -z "$(openstack server list)"
```
niemeyer, line 35: We have a system with very nice abstraction for composing test scenarios.  Is there a good reason for us to choose to cook all of them as a shell script inside a single task instead of using that composition to at least bundle closely related ideas together?

tests/openstack/task.yaml
```diff
@@ -0,0 +1,101 @@
...
+    grep 'cannot find ready marker in console output for .*: timeout reached' task.out
+
+    # Check that the server is discarded during the spread execution
+    for i in $(seq 10); do
+        if [ -z "$(openstack server list)" ]; then
+            break
+        fi
+        sleep 1
+    done
+    test -z "$(openstack server list)"
```
niemeyer, line 30: Is there a prior check that ensures that the list started empty in the first place?  Also, please consider the comment above in this context.

Also, won't OpenStack show the any used servers as terminated instead of just showing an empty list?

tests/openstack/task.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+    sudo -u stack -i /opt/stack/devstack/unstack.sh
+    sudo -u stack -i /opt/stack/devstack/stack.sh
+
+execute: |
+    # Check the instances which cannot be accessed are discarded
+    # The instance was created and the status has to be active to
+    # fail trying to access through ssh
+    spread openstack:cirros-64: -v -reuse -resend &> task.out || true
+    grep 'cannot find ready marker in console output for .*: timeout reached' task.out
```
niemeyer, line 21: This is not a great method to verify the contents, because on failure we'll get nothing, and won't know why we got nothing.  I believe we have common practices for this in the Spread world.  How do they look like?

tests/openstack/task.yaml
```diff
@@ -0,0 +1,101 @@
...
+    OS_PASSWORD=wrong spread openstack:cirros-64: -v -reuse -resend &> task.out  || true
+    grep 'cannot authenticate:.*returned unexpected status: 401' task.out
+    OS_USERNAME=wrong spread openstack:cirros-64: -v -reuse -resend &> task.out  || true
+    grep 'cannot authenticate:.*returned unexpected status: 401' task.out
+    OS_IDENTITY_API_VERSION=2 spread openstack:cirros-64: -v -reuse -resend &> task.out  || true
+    grep -E 'cannot authenticate:.*returned unexpected status: 404' task.out
+
+    # trigger 1 instance and check it can be listed and the garbage collect works
+    test "0" = "$(spread -gc | grep -c "Checking openstack instance")"
+    spread openstack:cirros-64: &>/dev/null &
```
niemeyer, line 64: When is this stopped/checked for correctness?

tests/openstack/task.yaml
```diff
@@ -0,0 +1,101 @@
...
+    # properties just can be set when the server is in ACTIVE status
+    SERVER_ID="$(openstack server list -f value | head -n1 | awk '{ print $1 }')"
+    test -n "$SERVER_ID"
+    for _ in $(seq 20); do
+        if openstack server show "$SERVER_ID" -f shell | MATCH 'status="ACTIVE"'; then
+            break
+        fi
+        sleep 1
+    done
+    openstack server show "$SERVER_ID" -f shell | MATCH 'status="ACTIVE"'
```
niemeyer, line 83: So all this test does is check that _some_ server has shown up?  Nothing else at all?

niemeyer (changes requested):
Thanks to everyone involved in getting this backend cooked. Let's see if we can get it merged in the near future.

A while ago I had already done an initial high-level pass on the logic with Michael, and still need to review it in more detail, but I'm not expecting major surprises there as my understanding is it's heavily based on the GCE backend.   So in this pass I reviewed mainly the bits surrounding the actual backend logic.   Once we get to some agreements on those I'll go in and do a more complete review on the backend details.

Please let me know how you'd like to proceed from here, otherwise.

.github/workflows/test.yaml
```diff
@@ -17,7 +17,7 @@ jobs:
 
       - name: Run tests
         run: |
-          spread google:
+          spread google:ubuntu-20.04-64: google:ubuntu-22.04-64-devstack:tests/openstack
```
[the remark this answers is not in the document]
sergiocazzolato: Hi, thanks for the review, 

In order to test openstack backend against a real openstack interface, we pre-configured a new image with devstack already installed. The main raeson was to speed up and simplify the test (so that image is only used to test openstack). Should I add this explanation in the workflow?

README.md
```diff
@@ -899,6 +900,66 @@ and need to be removed by hand.
...
+                networks:
+                    - network_external
+                    - network_pvn
+                groups:
+                    - group_external
+
+```
+
+The Openstack backend gets all the information to authenticate from the
+environment variables. The following variables have to be set:
```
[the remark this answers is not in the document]
sergiocazzolato: The most common mechanism  to connect a client through the Openstack API is by sourcing a file with teh environment variables (like the one we get to use openstack client in canonistack). For example in PS5, in the environment I also see the openstack env vars in my environment (they are managed by vault tool), so I presume juju is using those to connect to openstack.

I agree with you this is not a good practice because this means we need to have an env var with the user and password. I'll research which workaround we could use.

README.md
sergiocazzolato: In google we use the env var SPREAD_GOOGLE_KEY which has a link to the file with the following data:
"type"
"project_id"
"private_key_id"
"private_key"
"client_email"
"client_id"
"auth_uri"
"token_uri"
"auth_provider_x509_cert_url"
"client_x509_cert_url"

I understand that the equivalent are 
OS_PROJECT_ID <-> project_id
OS_AUTH_URL <-> auth_uri

spread.yaml
sergiocazzolato: removed

spread/export_openstack_test.go
sergiocazzolato: done

spread/project.go
sergiocazzolato: done

spread/runner.go
```diff
@@ -88,6 +88,8 @@ func Start(project *Project, options *Options) (*Runner, error) {
 			r.providers[bname] = AdHoc(project, backend, options)
 		case "humbox":
 			r.providers[bname] = Humbox(project, backend, options)
+		case "openstack":
```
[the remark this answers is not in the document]
sergiocazzolato: done

tests/openstack/spread.yaml
sergiocazzolato: done

tests/openstack/task.yaml
sergiocazzolato: devstack is preinstalled in the image, so when we start the tests we just prepare it.

The alternative is to download/configure and start devstack in prepare section, but it makes the test slower.

tests/openstack/task.yaml
sergiocazzolato: Added the initial check to verify there is not any server running. 

The command `openstack server list` doesn't show terminated servers, it just includes active ones.

tests/openstack/task.yaml
sergiocazzolato: Initially I though this but the problem I found is the time that devstack consumes to start (about 10 minutes), so if I create variants it will require more machines to get results. If makes sense I could move the scenarios to different variants and run them in parallel using more workers.

README.md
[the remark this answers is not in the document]
sergiocazzolato: I updated the documentation to explain better which env vars need to be defined. It is also supported to authenticate by using a key (similar to what we have in google). The key has to be stored in an env var to be loaded.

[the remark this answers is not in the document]
sergiocazzolato: An improvement could be to load the vars from a file (as we hav in google) instead of the env. @niemeyer What do you think?

spread/export_openstack_test.go
```diff
@@ -0,0 +1,89 @@
...
+	gooseClient "github.com/go-goose/goose/v5/client"
+	"github.com/go-goose/goose/v5/glance"
+)
+
+var (
+	OpenstackName = openstackName
+)
+
+type (
+	OpenstackProvider   = openstackProvider
```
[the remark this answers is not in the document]
ZeyadYasser: Good point, Thank you!

I refactored the tests to avoid the need to expose internal types.

spread/export_openstack_test.go
ZeyadYasser: Updated, Thank you!

spread/export_openstack_test.go
```diff
@@ -0,0 +1,89 @@
...
+func (opst *OpenstackProvider) FindImage(name string) (*glance.ImageDetail, error) {
+	return opst.findImage(name)
+}
+
+func (opst *openstackProvider) WaitProvision(ctx context.Context, serverData OpenstackServerData) error {
+	server := &openstackServer{
+		p: opst,
+		d: serverData,
+	}
+	return opst.waitProvision(ctx, server)
```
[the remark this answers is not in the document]
ZeyadYasser: Good point, Thank you!

I updated to use a simpler signature only exposing params relevant to the tests:
- `OpenstackWaitProvision(p Provider, ctx context.Context, serverID, serverName string)`
- `OpenstackWaitServerBoot(p Provider, ctx context.Context, serverID, serverName string, serverNetworks []string)`

README.md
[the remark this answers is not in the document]
sergiocazzolato: Finally I added a similar approach that the used in google.

niemeyer:
Folks, please schedule a call for us next week so we can have a general conversation about the approach for authentication and testing before we all spend too many cycles on different avenues.

spread/openstack.go
```diff
@@ -0,0 +1,787 @@
...
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	gooseClient "github.com/go-goose/goose/v5/client"
```
ZeyadYasser, line 13:
```suggestion
	gooseclient "github.com/go-goose/goose/v5/client"
```

spread/openstack.go
```diff
@@ -0,0 +1,787 @@
...
+
+func (s *openstackServer) Discard(ctx context.Context) error {
+	return s.p.removeMachine(ctx, s)
+}
+
+const openstackCloudInitScript = `
+#cloud-config
+runcmd:
+  - echo root:%s | chpasswd
+  - sed -i 's/^\s*#\?\s*\(PermitRootLogin\|PasswordAuthentication\)\>.*/\1 yes/' /etc/ssh/sshd_config
```
ZeyadYasser, line 162: it might be a good idea to comment out existing conflicting configs first, similar to what is suggested here https://github.com/snapcore/spread/pull/155#discussion_r1530108525

spread/openstack.go
```diff
@@ -0,0 +1,787 @@
...
+  - test -d /etc/ssh/sshd_config.d && echo 'PermitRootLogin=yes' > /etc/ssh/sshd_config.d/00-spread.conf
+  - test -d /etc/ssh/sshd_config.d && echo 'PasswordAuthentication=yes' >> /etc/ssh/sshd_config.d/00-spread.conf
```
ZeyadYasser, line 164: tiny nitpick
```suggestion
  - test -d /etc/ssh/sshd_config.d && echo -e 'PermitRootLogin=yes\nPasswordAuthentication=yes' > /etc/ssh/sshd_config.d/00-spread.conf
```

ZeyadYasser (commented):
Thank you, small comments

ZeyadYasser:
I dropped the spread test for openstack due to issues and inconsistencies faced with devstack where it cannot replicate a normal openstack cluster.

A better alternative is to do something similar to google. After the openstack backend is merged, we add it as a backend in [spread.yaml](https://github.com/snapcore/spread/blob/master/spread.yaml).

spread/openstack.go
```diff
@@ -0,0 +1,1030 @@
...
+	}
+
+	var err error
+
+	if err == nil && p.computeClient == nil {
+
+		// Load environment variables used to authenticate
+		if p.backend.Key != "" {
+			godotenv.Load(p.backend.Key)
+		}
```
cmatsuoka, line 978: Here we're getting a file defining environment variables and reading it into the environment (to be read from the environment later). could be nicer to extract the variables from the file and consume them directly (so they won't be accessible from the environment), or source the file before running spread and maybe dropping this field altogether?

spread/openstack.go
```diff
@@ -0,0 +1,1030 @@
...
+		cred, err := identity.CompleteCredentialsFromEnv()
+		if err != nil {
+			return &FatalError{fmt.Errorf("cannot retrieve credentials from env: %v", err)}
+		}
+
+		// Select the appropriate authentication method
+		var authmode identity.AuthMode
+		if os.Getenv("OS_ACCESS_KEY") != "" && os.Getenv("OS_SECRET_KEY") != "" {
+			authmode = identity.AuthKeyPair
+		} else if os.Getenv("OS_USERNAME") != "" && os.Getenv("OS_PASSWORD") != "" {
```
cmatsuoka, line 990: Maybe we could read the variables from either the environment or the file and keep an internal state, instead of using the environment as a data registry.

spread/openstack.go
```diff
@@ -0,0 +1,1030 @@
...
+			authmode = identity.AuthUserPassV3
+			if cred.Version > 0 && cred.Version != 3 {
+				authmode = identity.AuthUserPass
+			}
+		} else {
+			return &FatalError{fmt.Errorf("cannot determine authentication method to use")}
+		}
```
cmatsuoka, line 997: This could be delegated to a small helper so we won't need to worry about different modes in our main logic.

spread/openstack.go
```diff
@@ -924,6 +922,72 @@ func (p *openstackProvider) GarbageCollect() error {
...
+	}
+	identityAPIVersion := 3
+
+	// The location entry contains project/region
+	var region, proj string
+	loc := strings.SplitN(p.backend.Location, "/", 2)
+	if len(loc) == 2 {
+		proj = loc[0]
+		region = loc[1]
+	}
```
dariuszd21, line 956: Is it the problem if that's not true and proj and region will be unset ?

spread/openstack.go
```diff
@@ -924,6 +922,72 @@ func (p *openstackProvider) GarbageCollect() error {
...
+		return p.authErr
+	}
+
+	var err error
+	if p.computeClient == nil {
+		err = p.authenticate()
+	}
+
+	p.authComplete = true
+	p.authErr = err
```
dariuszd21, line 939: Should you still override `authComplete` and `err` if `p.computeClient != nil` ?

dariuszd21 (approved):
Looks good to me, just minor questions

spread/openstack.go
cmatsuoka: Changed to returning an error in this case.

spread/openstack.go
cmatsuoka: That is the pattern used in other backends to avoid repeating the check (presumably for `Allocate` and `GarbageCollect`?)

README.md
```diff
@@ -25,6 +25,7 @@ Spread
 [LXD backend](#lxd)  
 [QEMU backend](#qemu)  
 [Google backend](#google)  
+[Openstack backend](#openstack)
```
niemeyer, line 28: The casing of the project name is incorrect, both in documentation and in the code itself. The project seems to call itself as _OpenStack_ instead of _OpenStack,_ so I would suggest we follow suit and do a `s/Openstack/OpenStack/` everywhere. Let's please keep _"openstack",_ though (not _"openStack"_).

niemeyer (approved):
Thank you very much to everyone involved in developing and polishing this backend, and thanks @cmatsuoka for polishing the final details for unblocking it.  With your latest changes, the shared part of these changes seem to follow the existing patterns, and any remaining problems are going to be specific to OpenStack itself, so let's go ahead and get this in.

There's one final detail to be done shortly in a follow up, which is fixing the casing, per comment below. But this is fine to be done as a separate PR so we don't wait even longer than we already did.

PR #175, opened 2023-10-13
google: mark cache as ready after it got populated

mvo5:
The googleImagesCache is currently not marked `ready`. This means while it is correctly build it is not used in subsequent calls to projectImages(). This commit set the cache to ready after it got populated.

The first commit fixes the issue, the second adds the test that shows the behavior. Happy to talk about the test more, it makes google a bit easier to test but maybe you prefer a different approach :)

### DIFF
--- a/spread/client_test.go
+++ b/spread/client_test.go
@@ -16,13 +16,13 @@ type clientSuite struct{}
 var _ = Suite(&clientSuite{})
 
 func (s *clientSuite) TestDialOnReboot(c *C) {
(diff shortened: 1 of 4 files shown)


cmatsuoka (approved):

niemeyer (changes requested):
Thanks for the PR. Claudio, can we please have a look at these details?

niemeyer (approved):
Thanks everyone.

mvo5:
Super nice to see this getting in, thank you so much @niemeyer and @cmatsuoka

PR #178, opened 2024-02-21
backends: LXD: don't hardcode eth0

thp-canonical:
With the ubuntu-22.04 image, the network interface seems to be sometimes named "enp5s0" instead of "eth0". To be compatible with both old and new releases, consider any non-loopback interface viable for retrieving an IP address.

### DIFF
--- a/spread/lxd.go
+++ b/spread/lxd.go
@@ -443,9 +443,14 @@ func (p *lxdProvider) address(name string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	for _, addr := range sjson.State.Network["eth0"].Addresses {
-		if addr.Family == "inet" && addr.Address != "" {
-			return addr.Address, nil
+	for intf, network := range sjson.State.Network {
+		if intf == "lo" {
+			continue
+		}
+		for _, addr := range network.Addresses {
+			if addr.Family == "inet" && addr.Address != "" {
+				return addr.Address, nil
+			}
 		}
 	}
 	return "", &lxdNoAddrError{name}
--- a/spread/lxd.go
+++ b/spread/lxd.go
@@ -443,11 +443,11 @@ func (p *lxdProvider) address(name string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	for intf, network := range sjson.State.Network {
-		if intf == "lo" {
+	for ifacename, ifaceconf := range sjson.State.Network {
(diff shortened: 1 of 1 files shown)


bboozzoo (approved):
LGTM. I have the exact same change in my branch.

bboozzoo:
ping @ZeyadYasser, please have a look

ZeyadYasser (commented):
Thank you for fixing this, would it be a good idea to add a check that the server is reachable by the host through this IP address (using an icmp ping or something similar) before selecting it to avoid a case where there are multiple addresses but one of them is not accessible by the host.

CC: @bboozzoo

bboozzoo:
Unless the configuration adds another interface explicitly, the container will have `lo` and another non-`lo` interface (the problem here is that the name was hardcoded), so I suppose we could have a check if there's at most 2 interfaces, or simply only 1 that isn't lo? I'm not sure the connectivity check is necessary, as we did not have one before and the thing will simply fail on the next step trying to connect to the VM. This is exactly what happens without this change.

thp-canonical:
> I suppose we could have a check if there's at most 2 interfaces, or simply only 1 that isn't lo?

I'm assuming here that if multiple non-`lo` network interfaces are configured, any network interface would work for connecting (the first one that matches and has an IPv4 address), so this isn't strictly incompatible with multiple interfaces.

Spread didn't have the restriction of "only a single non-`lo` interface" before (it would work just fine with multiple interfaces, as long as there also was `eth0`), so I didn't want to introduce this additional restriction here.

If there are multiple interfaces where only some of them are suitable for Spread to connect via SSH, we probably need additional configuration anyway (and that is probably outside of the scope of this PR).

ZeyadYasser (approved):
LGTM, Thank you!

spread/lxd.go
```diff
@@ -443,9 +443,14 @@ func (p *lxdProvider) address(name string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	for _, addr := range sjson.State.Network["eth0"].Addresses {
-		if addr.Family == "inet" && addr.Address != "" {
-			return addr.Address, nil
+	for intf, network := range sjson.State.Network {
+		if intf == "lo" {
```
cmatsuoka, line 447: Could you change these variable names to match the corresponding chunk in #185?

cmatsuoka (approved):
Thank you

PR #180, opened 2024-02-21
fix: runner: ignore unknown file types and sockets (fixes #176)

### LINKED ISSUE #176: Packing project content fails if there are UNIX domain sockets
If there's any stray UNIX domain sockets in the project directory, Spread fails to pack the project.

To reproduce this issue, create a socket (you can Ctrl+C the command as soon as it's started, we just want the side effect of creating a socket):

```
nc -lkU foo.sock
```

With this in place, trying to run `spread` fails (example project has a LXD backend, but this issue is most likely independent of backend):

```
spread lxd -vv
```

The error message:

```
2024-02-05 15:15:39 Error packing project content for delivery: cannot pack project tree: tar: foo.sock: socket ignored
```

As a workaround, we currently run `find . -type s -delete` before running `spread` to remove any UNIX domain sockets.

GNU tar supports configurable warning messages using `--warning=` with:

```
              file‐ignored
                     "%s: Unknown file type; file ignored"
                     "%s: socket ignored"
                     "%s: door ignored"
```

The `file-ignored` warning ignores:

 * Unknown file types (see calls to `unknown_file_error()` in `src/create.c` in tar's source)
 * UNIX domain sockets
 * [Doors](https://en.wikipedia.org/wiki/Doors_(computing)), a Solaris-specific concept

It seems like one could add `--warning=no-file-ignored` to work around this issue:

```patch
diff --git a/spread/runner.go b/spread/runner.go
index c84ff63..ebf4b00 100644
--- a/spread/runner.go
+++ b/spread/runner.go
@@ -277,7 +277,7 @@ func (r *Runner) prepareContent() (err error) {
                return fmt.Errorf("cannot remove temporary content file: %v", err)
        }
 
-       args := []string{"c", "--exclude=.spread-reuse.*"}
+       args := []string{"c", "--exclude=.spread-reuse.*", "--warning=no-file-ignored"}
        if r.project.Repack == "" {
                args[0] = "cz"
        }
```

In my local tests, this fixed the issue and lets `spread` pack the local tree, ignoring any UNIX domain sockets.

thp-canonical:
Do not pack unknown file types + UNIX domain sockets when creating the tarball.

See #176 for a detailed description of the problem.

### DIFF
--- a/spread/runner.go
+++ b/spread/runner.go
@@ -277,7 +277,7 @@ func (r *Runner) prepareContent() (err error) {
 		return fmt.Errorf("cannot remove temporary content file: %v", err)
 	}
 
-	args := []string{"c", "--exclude=.spread-reuse.*"}
+	args := []string{"c", "--exclude=.spread-reuse.*", "--warning=no-file-ignored"}
 	if r.project.Repack == "" {
 		args[0] = "cz"
 	}


ZeyadYasser:
I replied in the issue https://github.com/snapcore/spread/issues/176#issuecomment-2009041848, I think this approach could accidentally suppress unexpected behaviors that are hard to debug. Would the exclude pattern suggested in the issue be a suitable alternative?

thp-canonical:
> I replied in the issue [#176 (comment)](https://github.com/snapcore/spread/issues/176#issuecomment-2009041848), I think this approach could accidentally suppress unexpected behaviors that are hard to debug. Would the exclude pattern suggested in the issue be a suitable alternative?

Agreed, I'll close this PR for now but keep the branch open in case we want to do something like this in the future. Let's continue the discussion in #176.

PR #182, opened 2024-02-26
tests(lxd): test newer releases

thp-canonical:
This will probably fail until #178, #179 and #181 are merged.

PR #185, opened 2024-03-27
spread: add LXD VM backend

bboozzoo:
Like LXD but with a twist, as instead of containers, it's using VMs. This is an improvement over the qemu backend, as LXD --vm generally runs with a virtio focused configuration ensuring that we're using hardware efficiently.


spread/lxd.go
```diff
@@ -443,9 +478,15 @@ func (p *lxdProvider) address(name string) (string, error) {
...
+	for ifacename, ifaceconf := range sjson.State.Network {
+		if ifacename == "lo" {
+			// ignore loopback interface
+			continue
+		}
+		for _, addr := range ifaceconf.Addresses {
+			if addr.Family == "inet" && addr.Address != "" {
+				return addr.Address, nil
+			}
 		}
```
thp-canonical, line 490: This one overlaps with #178, right? If this gets merged first, no need to merge #178.

spread/lxd.go
```diff
@@ -19,13 +19,28 @@ import (
 )
 
 func LXD(p *Project, b *Backend, o *Options) Provider {
-	return &lxdProvider{p, b, o}
+	return &lxdProvider{
+		project: p,
+		backend: b,
+		options: o,
```
thp-canonical, line 25: Should this set `vm: false` (even though that's the default) just for clarity?

spread/lxd.go
bboozzoo: yup, I think I mentioned somewhere that I had exact same change :) I will rebase as needed

spread/lxd.go
```diff
@@ -111,6 +126,13 @@ func (p *lxdProvider) Allocate(ctx context.Context, system *System) (Server, err
 	if !p.options.Reuse {
 		args = append(args, "--ephemeral")
 	}
+	if p.vm {
+		args = append(args, "--vm")
+	}
+	if p.backend.Memory > 0 {
+		mem := int(p.backend.Memory / mb)
+		args = append(args, "-c", fmt.Sprintf("limits.memory=%dMiB", mem))
+	}
```
thp-canonical, line 135: Does this work for both the VM and not VM backends? Probably worth mentioning in `README.md`.

spread/runner.go
```diff
@@ -82,6 +82,8 @@ func Start(project *Project, options *Options) (*Runner, error) {
 			r.providers[bname] = Linode(project, backend, options)
 		case "lxd":
 			r.providers[bname] = LXD(project, backend, options)
+		case "lxd-vm":
+			r.providers[bname] = LXDVM(project, backend, options)
```
thp-canonical, line 86: Can you document the new backend in `README.md`?

spread/lxd.go
bboozzoo: Yes, this applies to both containers and VMs. There's also CPU limit which can be set through `limits.cpu`, but that would require a field to be added to `Backend`. OTOH the QEMU backend could also support that with ease by passing `-smp <n>`, so maybe it's worth it.

spread/lxd.go
```diff
@@ -292,8 +320,15 @@ func (p *lxdProvider) lxdLocalImage(system *System) (string, error) {
 		return "", err
 	}
 
+	// TODO use lxd image list --format=json to get all of this in one call
```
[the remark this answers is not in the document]
bboozzoo: filed https://github.com/snapcore/spread/issues/186

spread/runner.go
bboozzoo: done

spread/lxd.go
bboozzoo: pushed an update to the README file

spread/lxd.go
bboozzoo: done

tests/lxd-vm/spread.yaml
```diff
@@ -0,0 +1,15 @@
+project: spread
+
+backends:
+    lxd-vm:
+        systems:
+            - ubuntu-20.04
```
thp-canonical, line 6: This should probably also test 16.04, 18.04 and 22.04 as per #182 (assuming there are VM images for all old releases as well)

bboozzoo: 16.04 and 18.04 don't work in a VM, lxd-agent does not come start, so I've limited this to 20.04 and 22.04 for now. Another issue I ran into, is that tests need to be run one by one, otherwise the host runs out of disk space.

bboozzoo: Eh, and 22.04 will only work when https://github.com/snapcore/spread/pull/184 or other similar fix lands.

cmatsuoka:
I'm seeing this message on execution using the lxd-vm backend, it can also be adjusted to mention virtual machine instead of container:
```
2024-08-27 18:07:46 Waiting for lxd container spread-8-ubuntu-22-04 to have an address...
```

spread/lxd.go
```diff
@@ -111,6 +127,16 @@ func (p *lxdProvider) Allocate(ctx context.Context, system *System) (Server, err
...
 	}
+	if p.vm {
+		args = append(args, "--vm")
+	}
+	if p.backend.Memory > 0 {
+		mem := int(p.backend.Memory / mb)
+		args = append(args, "-c", fmt.Sprintf("limits.memory=%dMiB", mem))
+	}
+	if system.Storage != Size(0) {
+		args = append(args, "-d", fmt.Sprintf("root,size=%d", system.Storage))
```
fnordahl, line 138: Containers could also benefit from support for the `storage` keyword.

spread/lxd.go
```diff
@@ -128,8 +154,18 @@ func (p *lxdProvider) Allocate(ctx context.Context, system *System) (Server, err
...
+	what := "container"
+	if p.vm {
+		what = "VM"
+	}
+	printf("Waiting for lxd %s %s to have an address...", what, name)
+	maxTimeout := 60 * time.Second
+	if p.vm {
+		// VM may take considerably longer to start
+		// TODO: should this be configurable?
+		maxTimeout = 180 * time.Second
```
fnordahl, line 166: I did not see this in my testing but it is probably true, are you using emulated VMs?

After the VM has an address, subsequent calls to `lxc exec` will fail until the `lxd-agent` inside the VM has started, https://github.com/canonical/spread/pull/221/commits/ab8232d56b5875f5e308ab91340e696a44e7a9ec could be useful for this.

spread/lxd.go
```diff
@@ -111,6 +127,16 @@ func (p *lxdProvider) Allocate(ctx context.Context, system *System) (Server, err
...
+	if p.vm {
+		args = append(args, "--vm")
+	}
+	if p.backend.Memory > 0 {
+		mem := int(p.backend.Memory / mb)
+		args = append(args, "-c", fmt.Sprintf("limits.memory=%dMiB", mem))
+	}
+	if system.Storage != Size(0) {
+		args = append(args, "-d", fmt.Sprintf("root,size=%d", system.Storage))
+	}
```
fnordahl, line 139: I see no way to specify the number of vCPUs for the VM. This is an important requirement as the default of 1 vCPU is most likely not sufficient.

In https://github.com/canonical/spread/pull/221/commits/804d1ee7fce005ba0c275bc24adc3318fe61ef7c generic support for the LXD instance type concept is introduced driven by the already existing `plan` keyword.

LXD applies these constraints both to containers and VMs.

fnordahl (commented):
Thank you for your work on this,I discovered this PR after authoring https://github.com/canonical/spread/pull/221

How can we work together to bring this to life?

yanksyoon:
Hello! Out of curiosity, any updates on this/how we might progress on it?

PR #190, opened 2024-11-29
Fix fetching artifacts when sudo -i changes directory

zyga:
When spread connects to a machine as a non-root user, then it injects calls to
sudo -i to run commands as root. This effectively makes the artifact sending
logic equivalent to this script:

```sh
cd directory
sudo -i tar ...
```

The problem is that sudo -i changes the current working directory back to
/root. Pass -C or --directory= to tar, so that all the files are referenced
relative to the task directory.


cmatsuoka (approved):

niemeyer (approved):
Thanks for the PR, zyga, and for the review, Claudio.

PR #202, opened 2025-01-03
feat: lxd spawn vm mode

yanksyoon:
LXD is launched as container. This is different in terms of behavior from all other modes (qemu, google, linode, ...etc).
This means that tests like docker are run via container in container methods which causes unexpected errors.
This PR proposes to launch LXD in VM mode not as containers to match the behavior.


bboozzoo:
See https://github.com/canonical/snapd/pull/14875 which adds more explicit support for LXD VMs.

PR #204, opened 2025-01-10
client: keep original PATH when executing local scripts

bboozzoo:
Keep the original PATH when executing local scripts, such that repack/adhoc snippets which reference tools installed in custom locations are able to access them.


zyga (approved):
I'm not sure why /snap/bin is there but it's already better than the current broken behaviour.

bboozzoo:
> I'm not sure why /snap/bin is there but it's already better than the current broken behaviour.

I'm guessing it's being added as a workaround for snapd (and thus /snap/bin) becoming available while the current shell is already executing a script that invokes spread. Most likely only relevant for some CI workflows, but I'd prefer to not break things for anyone and keep backward compatibility.

PR #206, opened 2025-01-20
project: allow changing the project root

cmatsuoka:
Add the reroot project attribute specifying the relative path from
the spread.yaml location to the actual project root.


spread/project.go
```diff
@@ -39,7 +39,8 @@ type Project struct {
 	Exclude []string
 	Rename  []string
 
-	Path string `yaml:"-"`
+	Path   string `yaml:"-"`
+	Subdir string `yaml:"project-subdir"`
```
niemeyer, line 43: project-dir?  Or even project-root?

Do we have any other analogous options at the moment?

The "subdir" bit feels a bit off considering that this is the topmost directory we have.

niemeyer (changes requested):
Looks easy and reasonable. Let's just sort out what's the proper wording for it.

spread/project.go
cmatsuoka: The existing similar case is `source-subir` in parts, but I'll be happy with `project-root` here.

niemeyer: As discussed, we have `path` already which represents the remote path, so we need to think about how to align both the terms used as well as to make the distinction between the local copy and the remote copy more obvious.

niemeyer: Per discussion, let's try `reroot` to see if it works well, in the same family of `repack` and `rename`.

cmatsuoka: Updated the attribute name to `reroot` and added a usage note to the documentation.

spread/project_test.go
```diff
@@ -78,19 +78,46 @@ suites:
...
+		errMsg   string // the expected error message
+	}{
+		{"spread.yaml", "", ""},
+		{".spread.yaml", "", ""},
+		{"other.yaml", "", "cannot load project file from .*: cannot find spread.yaml or .spread.yaml"},
+		{"spread.yaml", "subdir1", "cannot list suite tests/: open .*: no such file or directory"},
+		{"spread.yaml", "subdir2", ""},
+		{"spread.yaml", ".", ""},
+		{"spread.yaml", "/", ""},
+	} {
```
niemeyer, line 93: The code above wasn't very nice to read due to the wrong indentation, and it's now getting worse due to the way the for loop is embedding a struct and a list in it.  Can we please reorganize things for it to look clean?  The table should be moved out of the test (typical condensed syntax we use), with a named struct, tests defined in the table using field names as usual, ("reroot" instead of "subdir" maybe?), etc.

Not sure if the function is already here, but we have a nice "reindent" function in Chisel to allow the embedded yaml to be properly indented with the rest of the code.  Let me know if you need a link.

niemeyer (changes requested):
Tests seem to be failing at the moment.

Also, the test is reading a bit unfriendly at the moment. Some concrete suggestions below.

cmatsuoka:
Current test failure is strange, it affects only the lxc test on 20.04. This started out of nowhere and tests were previously passing, I'm investigating what could be happening there.

spread/project_test.go
```diff
@@ -62,35 +63,96 @@ type projectSuite struct{}
...
+	errMsg   string // the expected error message
+}
+
+var loadTests = []loadTest{
+	{
+		filename: "spread.yaml",
+		reroot:   "",
+		errMsg:   "",
+	},
+	{
```
niemeyer, line 78: Thanks, the test looks a lot more readable indeed.  Just one more detail: there's a minor twist here that we use in most projects to optimize the use of space, both horizontal and vertical:
```
var loadTests = []loadTest{{
	filename: "spread.yaml",
	reroot:   "",
	errMsg:   "",
}, {
	...
}, {
	...
}}
```
You can actually see that style in use at the bottom of the diff, on the tests for reindent.

testutil/reindent.go
```diff
@@ -0,0 +1,63 @@
+package testutil
```
niemeyer, line 1: These files should probably go to `spread/testutil/` instead of under the root directory.  We might also use `spread/internal/testutil`, but this is really a tool and we haven't made any promises that people can import the code as a library and expect it to work across versions.

niemeyer (changes requested):
Thank you. Last pass, and LGTM with these details sorted and tests passing.

cmatsuoka:
The CI failures with 20.04 (and sometimes 22.04) are caused by timeouts in lxd IP address acquisition. The timeout value is generous and should be enough under any normal circumstance. I'm increasing the value temporarily to unblock incoming PRs while I investigate this issue further with the administrators of the self-hosted runners.

niemeyer (approved):
Thanks for the tuning. LGTM

PR #210, opened 2025-02-05
chore: don't import deprecated io/ioutil package

cmatsuoka:
Package io/ioutil was deprecated in go 1.16. The same functionality
is provided by packages io or os, so replace the calls accordingly.


niemeyer (approved):
Nice, thanks Claudio.

PR #211, opened 2025-02-05
chore: don't import deprecated crypto/ssh/terminal

cmatsuoka:
Package golang.org/x/crypo/ssh/terminal moved to golang.org/x/term.
Using the previous package name is deprecated.


niemeyer (approved):
Keep them coming.  :)

PR #212, opened 2025-02-06
Added project-level artifacts

maykathm:
By introducing artifacts at the project level, spread can permit users to collect information about task runs and extract that information as a project artifact. So as the task artifacts serve to extract data about an individual spread task, the project artifacts serve to extract data about the entire project. As an example, if I wish to execute a dynamic code analysis during spread tasks, with project artifacts I can do so without causing changes to the individual tasks, but rather introduce my additional metrics collector at the prepare and restore layer.


maykathm:
Updated with changes requested in https://github.com/sergiocazzolato/spread/pull/4

zyga:
I'd love to see this land!

PR #231, opened 2025-04-14
refactor: adjust openstack name case

cmatsuoka:
Adjust the OpenStack name casing consistently across documentation
and code to match official spelling, but keep "openstack" in unexported
names (don't use mixed-case "openStack").


niemeyer (approved):
Thanks!

PR #244, opened 2025-06-30
Add support for keyboard-interactive authentication

zyga:
This makes spread able to authenticate to systems that do not allow plaintext passwords but DO allow keyboard-interactive authentication. Some newer systems use such method as there are some PAM interactions that prevent plaintext passwords from working (Gentoo).


spread/client.go
```diff
@@ -37,8 +37,20 @@ type Client struct {
...
-		Auth:            []ssh.AuthMethod{ssh.Password(password)},
+		User: username,
+		Auth: []ssh.AuthMethod{
+			ssh.Password(password),
+			ssh.KeyboardInteractive(func(user, instruction string, questions []string, echos []bool) ([]string, error) {
+				answers := make([]string, len(questions))
+				for i, _ := range questions {
+					if questions[i] == "Password: " {
+						answers[i] = password
+					}
```
bboozzoo, line 48: is this part driven by PAM?

zyga: I presume it is. Gentoo ships those override files in their images:

```
root@gentoo /etc/ssh/sshd_config.d # ls -la
total 12
drwxr-xr-x 2 root root 128 Jun 22 23:57 .
drwxr-xr-x 4 root root 291 Jul  1 07:17 ..
lrwxrwxrwx 1 root root  61 Jun 22 23:56 20-systemd-userdb.conf -> ../../../usr/lib/systemd/sshd_config.d/20-systemd-userdb.conf
-rw------- 1 root root 316 Jun  1 19:22 9999999gentoo.conf
-rw------- 1 root root 133 Jun  1 19:22 9999999gentoo-pam.conf
-rw------- 1 root root  79 Jun  1 19:22 9999999gentoo-subsystem.conf
```

```
root@gentoo /etc/ssh/sshd_config.d # cat 9999999gentoo*
# Allow client to pass locale environment variables (bug #367017)
AcceptEnv LANG LC_ALL LC_COLLATE LC_CTYPE LC_MESSAGES LC_MONETARY LC_NUMERIC LC_TIME LANGUAGE LC_ADDRESS LC_IDENTIFICATION LC_MEASUREMENT LC_NAME LC_PAPER LC_TELEPHONE

# Allow client to pass COLORTERM to match TERM (bug #658540)
AcceptEnv COLORTERM
UsePAM yes
# This interferes with PAM.
PasswordAuthentication no
# PAM can do its own handling of MOTD.
PrintMotd no
PrintLastLog no
# override default of no subsystems
Subsystem	sftp	/usr/lib64/misc/sftp-server
```

PR #258, opened 2025-09-17
test: build local spread binary

cmatsuoka:
Don't rely on GitHub runners providing a (potentially outdated)
spread binary. Instead, check out a reference version and build a
local binary to be used in the github CI test.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   run-tests:
-    runs-on: [self-hosted]
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -11,13 +11,35 @@ on:
...
+          sudo install -m755 spread-testing /usr/bin/spread
+          cd ..
+          rm -Rf spread-testing
+
+      - name: Run spread tests
+        run: |
+          uname -a
+          trap 'rm -f /tmp/key.json' EXIT
+          (umask 177; echo "${{ secrets.SPREAD_GOOGLE_KEY }}" | base64 -d > /tmp/key.json)
+          /usr/bin/spread google:
```
niemeyer: This is curious.. we're moving from keeping the key in memory in an environment variable, to keeping it in the filesystem, but the only place we use it seems to be right next to where we write down. I'm happy to merge this regardless so you can test it, if you wish, but why aren't we just doing what we used to do, which is reading the variable from the environment itself?

PR #287, opened 2026-04-14
ci: update pinned spread version

cmatsuoka:
Update the pinned spread version used in CI testing. This version
accepts authorized user credentials in the Google backend when
using string keys.


holmanb (approved):
This pins the workflow to the current latest commit, which I have tested locally.

niemeyer (approved):
Ack.

